PR #42, opened 2021-10-07
Right navigation fixes

imjuangarcia:
* Right-side navigation should show when in largest media query width;
* Right-side navigation should only show if there are > 1 items;
* Right-side navigation style fixes


tvst:
STATUS: Waiting on @randyzwitch's potential changes to using #anchor tags in left sidebar

snehankekre:
@tvst Anchor tags were removed by https://github.com/streamlit/docs/pull/117

randyzwitch:
We should probably rebase this, just to make sure this does what we're expecting

tvst (approved):

PR #69, opened 2021-10-13
Merge docs app

arnaudmiribel:
## What is that?

I created this [Docs app](https://share.streamlit.io/streamlit/docs/arnaud/app.py) in order to visualize how code blocks extracted from the docs would render, if we executed them. 

This enables to quickly spot typos, formatting problems although some exceptions raised are intentional.

## Why merge?

Merging this to branch `main` would enable the app to be as fresh as possible with the latest `.md` docs.


arnaudmiribel:
Sure @randyzwitch that makes sense. What's the best way for a Streamlit app to read (Markdown) files from a higher directory though?

PR #115, opened 2021-10-18
Makes API version selector respect dark-mode

### LINKED ISSUE #114: Darkmode doesn't work on drop down's text.
Hi, as the title suggests the drop down text isn't visible due to it having light background instead of a dark one.

Screenshot from 2021-10-16 20-24-23

snehankekre:
- Makes API version selector drop-down respect dark-mode by setting `background-color: $off-black;`.
- Fixes #114.


randyzwitch:
@imjuangarcia if you're good with this, I am.

PR #119, opened 2021-10-18
Create LICENSE

### LINKED ISSUE #113: Add LICENSE file
@tvst I assume we'd use Apache 2.0 here as well?

randyzwitch:
Fixes #113


randyzwitch:
Making the executive decision to go with it, can't imagine why we'd do something different than the original docs (which were by default Apache 2.0 since they were in the larger streamlit repo)

PR #124, opened 2021-10-20
Fix on focus ring for section headings

imjuangarcia:
This PR adds focus ring to section headings, such as `Streamlit Library`, `Streamlit Cloud` as `Knowledge Base`, so you can tab through it

### DIFF
--- a/styles/components/navigation/sideBar.scss
+++ b/styles/components/navigation/sideBar.scss
@@ -173,16 +173,16 @@
             }
         }
 
-        a:focus:not(:focus-visible) {
-          outline: none;
-        }
-
         ul, ol {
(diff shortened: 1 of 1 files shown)


styles/components/navigation/sideBar.scss
```diff
@@ -173,17 +173,18 @@
             }
         }
 
-        a:focus:not(:focus-visible) {
```
tvst, line 176: I don't think you should remove this.

If you remove it, then the focus ring appears when you click on a navbar item, which is no good.

tvst (changes requested):

tvst (approved):

PR #154, opened 2021-11-04
Make function signature in docstrings *not* include type annotations.

tvst:
For this, I had to write my own function that generates signatures by inspecting functions.

This PR also pins our Dockerfile to Python 3.9 because otherwise we only generate JSON data for a handful of Streamlit versions (i.e. only the ones that support 3.10+).


snehankekre (approved):
This is brilliant! LGTM @tvst

imjuangarcia (commented):
This is behind my area of expertise, so I can't really say much about it. However, I tested the Deploy Preview and all works as expected, so feel free to merge if all looks good on your end!

PR #155, opened 2021-11-08
Add long descriptions to API reference docstrings

snehankekre:
### Adds long descriptions / extended summaries to API reference docstrings

Currently, we pull a one-line (short) description of each Streamlit function in the API reference. However, a number of Streamlit functions include long descriptions / extended summaries in reST format that span multiple lines. These should be pulled, parsed, and included in the docstring descriptions. 

Here is an example of a long description included in the `st.form` docstring source, how we currently (don't) display it, and how changes from this PR fix it:

### Docstring source:
image

### Current
image

### After PR
image


randyzwitch (approved):
I approve, but it's not clear what the .json difference @snehankekre since it doesn't show in the interface. Do you know what the problem might be?

snehankekre:
@randyzwitch The conflict was a result of merging the autoformat PR. I've resolved the conflict.

PR #110, opened 2021-10-15
WIP: Fix on versioned urls

imjuangarcia:
This PR fixes the issue with the version number getting lost when you click on an unversioned page by preserving the version number in `localStorage`.

PR #158, opened 2021-11-08
Use custom hook to hold the version number selected

imjuangarcia:
This is a proof of concept PR, with an alternative implementation to fix the issue with the version number getting lost when you click on an unversioned page.

## The problem

If you go to a versioned page (for example [here](https://docs.streamlit.io/library/api-reference/write-magic/st.write)), change the version from the selector, then go to an unversioned page (such as [this one](https://docs.streamlit.io/library/get-started)) and then return to the versioned page, you lose the version you previously picked.

## The first solution

[This previous PR](https://github.com/streamlit/docs/pull/110) approached the problem by preserving the version number in `localStorage`, but it could lead to issues with users changing the version number, leaving the page, coming back and not realizing they have an older version being retrieved automatically from `localStorage`.

## The new solution

This PR approaches this problem by preserving the version number in local state, and calling a custom `useVersion` hook to get the version number from the URL.

## Know issues

A similar issue to what [had been discussed here](https://github.com/streamlit/docs/pull/110#issuecomment-945899570) is still happening. See video reproduction [here](https://share.getcloudapp.com/z8u11b8Y). Steps to reproduce:

* Navigate to the `st.write()` API reference page;
* Change the version number to an older version;
* Go to any unversioned page, such as the Library landing page;
* Return to the `st.write()` page using the sidebar;
* Navigate to other versioned URL, such as the `st.button()` page.

You should see when you navigate back to the `st.write()` page, that the page shows the older version content, but the older version number isn't showing up on the URL. Only after you click on the page, and navigate to `st.button()` the URLs get updated with the older version number on them.


tvst:
@imjuangarcia I just pushed some changes which I think fixes all the remaining bugs.

What this does:
* Adds a new property to menu items, called `isVersioned`. If true, this is a menu item (and a page) that accepts versions in the URL. Previously, in `NavItem` we were searching through the`paths` variable to figure out if the page accepted versions, but in some cases `paths` was not present, which caused the bug you described.
  Now we no longer need `paths` anywhere in the sidebar.
* Stops using the router to figure out the current version since we already have that information from the static props in `[...slug].js`.
* Moves the version state into a React context. This allows pages that don't use `[...slug].s` to also remember the version.

imjuangarcia:
@tvst Looks and works great! Want to merge it on `main`, or do you want me to do it?

tvst:
Go for it!

tvst (approved):

imjuangarcia:
@tvst Mind doing a last sanity check on this PR? I had to merge it with `main`, and several components changed pretty drastically due to the refactoring.

I took your changes, and then refactored those components again on 65c3e07. Everything seems to be working as expected, but a final green light would be desirable. Thanks!

tvst (approved):
Looks good. I'm not seeing any problems...

PR #162, opened 2021-11-09
Convert `/utilities` folder to functional components

imjuangarcia:
This PR starts refactoring the components under the `/utilities` folder, from class-based to functional ones.

This PR shouldn't introduce any changes in functionality, but please pay special attention to the `floatingNav` and `helpful` components, as those involved rather big changes in the code.


components/utilities/floatingNav.js
```diff
@@ -1,204 +1,145 @@
-import findIndex from "lodash/findIndex";
-import pullAt from "lodash/pullAt";
-import React from "react";
-import { breadcrumbsForSlug } from "../../lib/utils.cjs";
-
+import React, { useState, useEffect, useRef } from "react";
```
tvst, line 1: Note to self: Juan asked for a close review of this code

components/utilities/helpful.js
```diff
@@ -1,64 +1,52 @@
-import React from "react";
+import React, { useState, useEffect, useRef } from "react";
```
tvst, line 1: Note to self: Juan asked for a close review of this code

components/utilities/floatingNav.js
```diff
@@ -1,204 +1,145 @@
...
-    window.removeEventListener("ChangeTheme", this.handleTheme);
-    this.props.router.events.off("routeChangeComplete", this.handleRouteChange);
-  }
+  return nestedHeadings;
+};
 
-  handleTheme() {
-    this.setState({ theme: document.body.dataset.theme });
-  }
+const useIntersectionObserver = (setActiveId) => {
```
tvst, line 43: I don't really understand what this function is supposed to do. Can you add a docstring explaining?

components/utilities/helpful.js
```diff
@@ -1,64 +1,52 @@
...
+    other: false,
+    helpful: true,
+    improvements: [],
+    notes: "",
+    improvementsString: "",
+    moreExamples: false,
+    clearerSteps: false,
+    moreInformation: false,
+    other: false,
+  });
```
tvst, line 20: Break into multiple states. (Although [it's ok to group values that change together](https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables))

components/utilities/helpful.js
```diff
@@ -70,152 +58,154 @@ export default class Helpful extends React.Component {
       pull(improvements, name);
     }
 
-    this.setState({ [name]: value });
-    this.setState({ improvements: improvements });
-    this.setState({ improvementsString: improvements.join(",") });
-  }
+    console.log(improvements);
```
tvst, line 61: Remove.

tvst (changes requested):

components/utilities/gdpr.js
```diff
@@ -12,7 +12,7 @@ const EUROPE_TZ_OFFSET_EAST = -3 * 60;
 
 const KEY = "InsertAnalyticsCode";
 
-export default function GDPRBanner(gdrp_data) {
+const GDPRBanner = (gdrp_data) => {
```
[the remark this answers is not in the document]
imjuangarcia: Fixed on d01c63d

components/utilities/helpful.js
imjuangarcia: Makes sense! Fixed on 18e3592, and kept the feedback values in one single state update per the link you recommended, LMKWYT!

components/utilities/search.js
```diff
@@ -12,89 +12,78 @@ import {
...
-      results: [],
-    };
+const Search = () => {
+  const [state, setState] = useState({
+    modal: false,
+    hotKey: "",
+    windowWidth: null,
+    indexFocus: 0,
+    results: [],
+  });
```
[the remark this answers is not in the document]
imjuangarcia: Fixed on 96ffedc

components/utilities/floatingNav.js
imjuangarcia: Makes total sense. Since this component was refactored almost entirely, I used a different logic to get the headings on the page and map them, but didn't properly explained it. Added the explanation on aa1e2c2

tvst:
Remove `components/.DS_Store`

imjuangarcia:
> Remove `components/.DS_Store`

Fixed on f969622. It's on the `.gitignore` file though, not sure how it snuck in.

components/utilities/floatingNav.js
```diff
@@ -1,204 +1,158 @@
...
+// This is the function that initializes the intersection observer, and attaches it to the elements we want to track, our page headings
+const useIntersectionObserver = (setActiveId) => {
+  // Since we want to keep track of the visibility of all the heading elements, we'll store these values in a useRef hook
+  const headingElementsRef = useRef({});
+
+  useEffect(() => {
+    // This is our callback function. The observer will call this function each time elements scroll in or out of view.
+    // When we first render the page, it calls the callback with a list of all the elements we want to keep track on the page. As elements scroll in and out of view, it will call the callback with these elements to update the visibility.
+    const callback = (headings) => {
+      headingElementsRef.current = headings.reduce((map, headingElement) => {
```
tvst, line 52: I think this would be cleaner/more understandable as a simple forEach:

```js
      headings.forEach((headingElement) => {
        const link = headingElement.target.getElementsByTagName("a");
        if (link.length > 0) {
          const target = link[0].getAttribute("href");
          headingElementsRef.current[target] = headingElement;
        }
      });
```

components/utilities/floatingNav.js
```diff
@@ -1,204 +1,158 @@
...
+        const link = headingElement.target.getElementsByTagName("a");
+        if (link.length > 0) {
+          const target = link[0].getAttribute("href");
+          map[target] = headingElement;
+        }
+        return map;
+      }, headingElementsRef.current);
```
tvst, line 59: Is it necessary to do this loop to grab heading refs _every time_ an elements scrolls into view? Can't we do this just once outside `callback`?

tvst (approved):

tvst (changes requested):
Ooops, didn't mean to approve just yet. I'm still looking at `floatingNav.js`

components/utilities/floatingNav.js
imjuangarcia: Fixed on d27b9aa

components/utilities/floatingNav.js
imjuangarcia: I tried to move the `forEach` loop to a separate function outside the Observer, and call it only when there are no heading references, but that won't work if we want to highlight the active title as we scroll.

I've been reading the [Intersection Observer docs](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver#syntax) and the callback function seems to be required for this matter. Are you concerned about memory leak issues? Or is this request aimed to make the code more efficient?

tvst:
Something broke in the current version. When you go from one page to another, the floatingNav remains the same.

tvst:
I worked on the floatingNav code to simplify it and avoid so many loops / recomputations. LMK what you think:

https://github.com/streamlit/docs/pull/184

imjuangarcia:
> I worked on the IntersectionObserver code in floatingNav.js to simplify it and avoid so many loops / recomputations

@tvst I replied on your PR, but putting the comment here for visibility as well: I like the cleanliness of the new code! The issue I see with this implementation though, is that it doesn't consider the scenario when you have multiple headings visible on the page. See [this example](https://share.getcloudapp.com/5zu9RYyw).

imjuangarcia:
> Something broke in the current version. When you go from one page to another, the floatingNav remains the same.

This issue is also related with changing the `<a>` tags for `<Link>` components on the sidebar. Since there aren't full page refreshes anymore, the floating nav wasn't updating properly. I just added a dependency on our `useEffect` calls to trigger the re-renders when the page changes.

imjuangarcia:
> I worked on the IntersectionObserver code in floatingNav.js to simplify it and avoid so many loops / recomputations.

I like its cleanliness! I merged with this PR here: 83f0a24 so it's easier to work on. What I'm not sure of is the fact that the active element is the bottom-most. That can lead to issues like [this one](https://share.getcloudapp.com/5zu9RYyw) when multiple headings are visible on the page. For example, in that screenshot, I clicked on `Virtual Private Cloud`, but the highlighted element is `Penetration testing`.

tvst (approved):

PR #165, opened 2021-11-11
Add meta descriptions to few API pages

snehankekre:
### Add meta descriptions to few API pages
- At @CharlyWargnier's request, this PR adds meta descriptions to five Streamlit API pages. The idea is to check whether search engines properly index these meta descriptions and, if so, add meta descriptions to the remaining Streamlit API pages in a future PR.
- The meta description is a small blurb that appears in search engines. It gives users the right context, helps them decide if they want to click, improves branding, UX, and click-through rates from search.

### DIFF
--- a/content/library/api/text/caption.md
+++ b/content/library/api/text/caption.md
@@ -1,6 +1,7 @@
 ---
 title: st.caption
 slug: /library/api-reference/text/st.caption
+description: st.caption displays text in small font.
 ---
 
 <Autofunction function="streamlit.caption" />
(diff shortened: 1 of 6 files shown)


randyzwitch:
There's no risk to doing all of them at once, is there? If the hypothesis is 'Will google take these?', if they don't, then its the status quo. If they do, all 65 get it that much faster.

CharlyWargnier:
There's no risk per se, although we may need to adjust the batch based on how indexation goes.
I'd recommend testing these 5 first. I will monitor Google's indices and will feedback next week.

snehankekre:
Thanks for providing the rationale, @CharlyWargnier. Merging this PR now.

PR #167, opened 2021-11-12
Fix spacing in frontend note

randyzwitch:
image

Adding a line break should fix this on the bi-directional components section


randyzwitch:
Fixed note

image

Issue #159, opened 2021-11-09
"App model" fragment in the "Main concepts" docs page seems broken


vdonato:
Navigating to https://docs.streamlit.io/library/get-started/main-concepts#app-model does not scroll the page down to the corresponding section. I tried navigating to a few other parts of the page via fragment, and they don't seem to be running into the same issue.

What seems to be a related bug is present when clicking on the page navigation links on the right hand side of the page -- clicking on the "App model" link brings you to the caching header.

snehankekre:
Thanks for filing this bug report @vdonato! We're working on a fix.

PR #169, opened 2021-11-12
Use remark-unwrap-images to prevent html errors

imjuangarcia:
This PR fixes the issue mentioned [here](https://github.com/streamlit/docs/issues/159).

## The problem

The root cause for this were a number of html violations generated by our [custom image component](https://github.com/streamlit/docs/blob/main/components/blocks/image.js): See [this screenshot](https://share.getcloudapp.com/5zu9PpAG) as an example. These errors were causing the hydration process to behave incorrectly.

Specifically, the root of the problem was the fact that the image is being wrapped by a `<p>` when it shouldn't be. See [this screenshot as an example](https://share.getcloudapp.com/GGu4zxKE).

## The solution

Using [`remark-unwrap-images`](https://github.com/remarkjs/remark-unwrap-images) when parsing the Markdown files, to remove the wrapping `<p>` tags.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,6 @@
         "node-html-parser": "^4.1.5",
         "nprogress": "^0.2.0",
         "prettier": "^2.4.1",
-        "pretty-quick": "^3.1.1",
         "prismjs": "^1.25.0",
         "react": "^17.0.2",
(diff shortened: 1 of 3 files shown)


snehankekre (approved):
LGTM! :heart:

PR #172, opened 2021-11-16
Sanity Checks - Update Streamlit version # to latest

mayagbarnes:
Came across [this page](https://docs.streamlit.io/knowledge-base/using-streamlit/sanity-checks) in docs and noticed old version number in Check 2. This PR updates it, but perhaps manually updating for each release isn't very practical?

Screen Shot 2021-11-16 at 10 09 31 AM

### DIFF
--- a/content/kb/using-streamlit/sanity-checks.md
+++ b/content/kb/using-streamlit/sanity-checks.md
@@ -39,7 +39,7 @@ pip install --upgrade streamlit
 streamlit version
 ```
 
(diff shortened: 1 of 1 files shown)


randyzwitch (approved):

randyzwitch:
I might be wrong, but I think we might've had a script that updated these in the past? @snehankekre would know. But yeah, something for us to keep up to date.

PR #173, opened 2021-11-17
Convert `/navigation` folder to functional components

imjuangarcia:
This PR refactors the components under the `/navigation` folder, from class-based to functional ones.

### DIFF
--- a/components/navigation/navItem.js
+++ b/components/navigation/navItem.js
@@ -11,18 +11,6 @@ const NavItem = ({
   version,
   maxVersion,
 }) => {
-  const trueName = () => {
-    return page.path.split("/").pop();
-  };
-
-  const cleanName = () => {
-    return trueName().replaceAll("-", " ");
-  };
-
-  const shouldHaveID = () => {
-    return offScreen ? page.name : false;
-  };
-
   let subNav;
   let navItem;
   let navBox;
--- a/components/navigation/navItem.js
+++ b/components/navigation/navItem.js
@@ -1,89 +1,87 @@
 import React from "react";
-
-import Link from "next/link";
 import { urlInChildren } from "../../lib/utils.cjs";
-
 import NavChild from "./navChild";
 
-export default class NavItem extends React.Component {
-  constructor(props) {
-    super(props);
-  }
+const NavItem = ({
+  page,
+  offScreen,
+  slug,
+  condensed,
+  paths,
+  version,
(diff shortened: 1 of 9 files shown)


components/navigation/navItem.js
```diff
@@ -1,89 +1,87 @@
...
+  offScreen,
+  slug,
+  condensed,
+  paths,
+  version,
+  maxVersion,
+}) => {
+  const trueName = () => {
+    return page.path.split("/").pop();
+  };
```
tvst, line 16: `trueName` could just be a constant, no?

```
const trueName = page.path.split("/").pop();
```

components/navigation/navItem.js
```diff
@@ -1,89 +1,87 @@
...
-    const props = this.props;
+  const cleanName = () => {
+    return trueName().replaceAll("-", " ");
+  };
 
-    let subNav;
-    let url_in_child = false;
+  const shouldHaveID = () => {
+    return offScreen ? page.name : false;
+  };
```
tvst, line 24: `shouldHaveId` could just be a constant, no?

```
const shouldHaveId = offScreen ? page.name : false;
```

components/navigation/sideBar.js
```diff
@@ -1,110 +1,96 @@
...
+  const [state, setState] = useState({
+    condensed: false,
+    loading: true,
+    depth: 1,
+    sticky: false,
+    over: false,
+    open: false,
+    theme: "light-mode",
+    menu,
+  });
```
tvst, line 15: I think it's more correct to pull each of these into their own `useState`:

```
const [isCondensed, setIsCondensed] = useState(false);
const [isLoading, setIsLoading] = useState(false);
...
```

tvst (approved):

components/navigation/navItem.js
imjuangarcia: In fact, these weren't being used, so I deleted them on 51f922b

components/navigation/navItem.js
imjuangarcia: In fact, these weren't being used, so I deleted them on 51f922b

components/navigation/sideBar.js
imjuangarcia: Makes sense! Added them on b18db14

tvst (approved):

Issue #175, opened 2021-11-17
Docs UI glitch between 1025px and 1249px screen width


StefanBrand:
## Problem

Between 1025px and 1249px screen width, the side bar overlays docs content, even if collapsed. Tested in Firefox 94.0 on my 1280x1024px screen (due to Ubuntu dash bar, actual browser window width is 1204px).

The screenshot demonstrates the behaviour. It is not possible to click links under the red area.
![image](https://user-images.githubusercontent.com/8948865/142172251-5e88edc2-37d9-4b0a-b17d-07c739138770.png)

## Solution
Fix media queries

## Workaround

Zoom out or make window width smaller than 1025px.

StefanBrand:
This still happens to me on https://docs.streamlit.io/library/changelog

snehankekre:
Thanks for reporting this @StefanBrand. Reopening this issue and looping in @imjuangarcia to take a look.

PR #179, opened 2021-11-22
Fix on sidebar links overlay on small desktop resolutions

imjuangarcia:
This PR fixes https://github.com/streamlit/docs/issues/175 by removing the higher `z-index` value when the sidebar is collapsed, and adding it when it's expanded


snehankekre (approved):
Amazed at the quick turnaround time! LGTM :smile:

PR #184, opened 2021-11-26
[NOT FOR MERGE] Convert utilities to functional components

tvst:
Juan: PTAL at `components/utilities/floatingNav.js`'s [IntersectionObserver](https://github.com/streamlit/docs/pull/184/files#diff-d8e99f884e0784bb9a2bbfb6abba0f60a530b30bc9a7d01c99ed2efaae9c4c0aR44), which I think is cleaner than in the original PR https://github.com/streamlit/docs/pull/162.

BTW, this branch does not include all the latest changes from your PR. So if you decide to use this, you'll need to merge. (I didn't merge because the current version of the original PR has a bug. See my comment over there.)


imjuangarcia:
@tvst Thank you for testing this out! I like the cleanliness of the code. The issue I see with this implementation though, is that it doesn't consider the scenario when you have multiple headings visible on the page. See [this example](https://share.getcloudapp.com/5zu9RYyw).

PR #185, opened 2021-11-26
Various fixes on Sidebar links

imjuangarcia:
# The Problem

When navigating the docs site using the sidebar, the `/library` URL should redirect folks to `/library/get-started`, but it isn't. Seems like using Next’s `<Link />` component on the sidebar is preventing the page from redirecting, since there’s no full page refresh.

# The Solution

This PR adds a condition to check the page URL, and if it's `/library`, we replace Next's `<Link />` component, and use an `<a>` tag instead.


tvst (approved):
I think in the past we used the menu system for this, no?

If not possible, then this LGTM.

PR #203, opened 2021-12-08
Add KB article on increasing file uploader limit

snehankekre:
This PR adds a KB article on how to increase the upload limit of `st.file_uploader` in apps deployed to Streamlit Cloud. The solution is to set the appropriate config option in a global config file and upload said file to the app's GitHub repo.

This article was inspired by this [forum thread](https://discuss.streamlit.io/t/streamlit-drag-and-drop-capping-at-200mb-need-workaround/19803/2).

### DIFF
--- a/content/kb/deployments/increase-upload-limit-cloud.md
+++ b/content/kb/deployments/increase-upload-limit-cloud.md
@@ -1,0 +1,46 @@
+---
+title: How do I increase the upload limit of st.file_uploader on Streamlit Cloud?
+slug: /knowledge-base/deploy/increase-file-uploader-limit-streamlit-cloud
+---
+
+# How do I increase the upload limit of st.file_uploader on Streamlit Cloud?
+
+## Overview
+
+By default, files uploaded using [`st.file_uploader()`](/library/api-reference/widgets/st.file_uploader) are limited to 200MB. You can configure this using the `server.maxUploadSize` config option.
+
+Streamlit provides [four different ways to set configuration options](/library/advanced-features/configuration#set-configuration-options):
+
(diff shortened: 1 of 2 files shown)


randyzwitch:
Feels like the answer could be #1 or #2 here...you could either make a `.streamlit` folder, or put a `config.toml` file at the top level of the repo. I wonder if Cloud has a preference on which to say.

snehankekre:
> Feels like the answer could be #1 or #2 here...you could either make a .streamlit folder, or put a config.toml file at the top level of the repo.

You're right! I completely overlooked option 2. I've seen repos that use option 2 as well. Will run it by Cloud and make changes to this PR.

snehankekre:
> you could either make a .streamlit folder, or put a config.toml file at the top level of the repo

#### When deploying to Streamlit Cloud

@randyzwitch AFAIK to be used, a `config.toml` file has to exist within a `.streamlit/` folder. That folder should exist at the top level (root) of the repo, regardless of the location of the app `app.py`. 

If the app is in a folder `dir/app.py`, the config file should be `.streamlit/config.toml`, not `dir/.streamlit/config.toml`. 

To verify, I created a this [repo](https://github.com/snehankekre/config-location-test) and [app](https://share.streamlit.io/snehankekre/config-location-test/main/dir/app.py):
-  `app.py` and `.streamlit/config.toml` are within `dir/`
- Notice how the app does not use the custom (dark) theme defined in the config. 
- It picks up the theme only when `.streamlit/config.toml` is moved out of `dir/` to the root of the repo.

Takeaway is that option 1 is still the correct answer. WDYT?

randyzwitch:
Thanks for confirming @snehankekre, #1 seems like the answer :)

PR #211, opened 2021-12-14
Fix docstrings: show return type info

snehankekre:
Fixes a bug [noticed by Johannes](https://www.notion.so/streamlit/Show-return-types-in-widget-docstrings-440762c59fc34005b88eb876b4de0178) whereby return type information in widget docstrings was not being displayed.

### Changes:
-  Updates  `python/generate.py` to obtain return type info from docstrings
- Creates a new table in `autofuncton.js` to display return type info
- Modifies `table.js` to conditionally render the new table:
  -  If docstring belongs to an input widget, display the new table with return type info
  - Else, don't display the table

### Revised
image

### Current
image

### DIFF
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -238,6 +238,7 @@ const Autofunction = ({
       }
       footRows={returns.length ? returns : null}
       additionalClass="full-width"
+      footers={footers}
     />
   );
 
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -98,7 +98,6 @@ const Autofunction = ({
   let functionDescription;
   let header;
   let body;
-  let returnBody;
 
   if (streamlitFunction in streamlit) {
     functionObject = streamlit[streamlitFunction];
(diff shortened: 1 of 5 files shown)


components/blocks/autofunction.js
```diff
@@ -214,6 +231,21 @@ const Autofunction = ({
       }
       rows={args.length ? args : null}
       addtionalClass="full-width"
+    />
+  );
+
+  returnBody = (
```
imjuangarcia, line 237: @snehankekre and @tvst: Do you think it makes sense to have the returns in a separate table? Or should it be part of the above one?

The reason I'm asking this is the fact it creates a [visual separation between the table contents](https://share.getcloudapp.com/qGurPjnz) and I'm not sure that's what we want

imjuangarcia (commented):
The JS piece of this PR looks good to me, but I have a small design nit/question that I'd like @tvst to chime in, so I'm leaving the approval to him just in case he has additional feedback!

components/blocks/autofunction.js
snehankekre: > Do you think it makes sense to have the returns in a separate table? Or should it be part of the above one?

@imjuangarcia No, it shouldn't be a separate table and should be a part of the above one. I should have mentioned above that I created a new table for the returns because I didn't know the JS needed to add it to the above table. :sweat_smile: 

My janky implementation should not be merged IMO. I'd love your and/or Thiago's help with adding the returns info to the original table (perhaps as an additional `<tbody></tbody>`). :point_left: This is what I didn't know how to do in JS.

imjuangarcia: @snehankekre Got it, thanks for letting me know! Just fixed this issue on a7a1616, using a `<tfoot>` element. Let me know what you think!

snehankekre: @imjuangarcia Looks splendid! :tada:  After adding the missing footers on 3b4fbb2, I think we're good to merge, unless @tvst or @randyzwitch have objections.

imjuangarcia (approved):

PR #213, opened 2021-12-15
Replace Segment tag ID

imjuangarcia:
Replace Segment tag ID to track events from the docs site properly. It seems like the docs’ analytics data isn’t writing to Segment properly (see below 👇)

Screen Shot 2021-12-14 at 12 14 06 PM.

I checked the JS code and it seems to be working as expected (the Segment tag is getting added to the DOM when allowing the cookie banner), but the tag ID seems to be incorrect (it was the one we use on the website). The tag is now replaced to be the one from Segment's source, and my tests seem to be getting logged on the debugger:

Screen Shot 2021-12-15 at 18 10 06

### DIFF
--- a/components/utilities/gdpr.js
+++ b/components/utilities/gdpr.js
@@ -139,9 +139,9 @@ function insertAnalytics() {
           n.parentNode.insertBefore(t, n);
           analytics._loadOptions = e;
         };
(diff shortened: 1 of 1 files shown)


snehankekre (approved):
LGTM!

PR #214, opened 2021-12-17
Adjustments on Search

imjuangarcia:
* Enhance hover effects;
* Fix on up/down/enter key combinations not working;
* Horizontally center the search icon

### DIFF
--- a/components/utilities/search.js
+++ b/components/utilities/search.js
@@ -100,6 +100,11 @@ const Search = () => {
   const goToResult = () => {
     let index = indexFocus;
 
+    if (index <= 0) {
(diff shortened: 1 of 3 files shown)


tvst (approved):
Thanks!

We should also make it possible to press Enter to go to the first result when *nothing* is selected.

This flow:
1. Cmd-K 
2. Type "button"
3. Press Enter

But that can be in another PR if you like

imjuangarcia:
> We should also make it possible to press Enter to go to the first result when _nothing_ is selected.

Fixed on c84d719

PR #216, opened 2021-12-20
Adds support to use Tailwind `dark` modifier

imjuangarcia:
Now that we're starting to move the docs' CSS to Tailwind, it would be good to use [Tailwind's `dark` modifier](https://tailwindcss.com/docs/dark-mode) to style things when on dark mode.

To do that without solely relying on the user's `media` preference, we need to append a `dark` class on the `<html>` tag. Currently, we also have a `data-theme` attribute on the `<body>`, which is used to style things using Sass. As we progressively move away from Sass and into Tailwind CSS we'll likely remove the data attribute, but for now, we have to support both scenarios until we've refactored the entire codebase.

### DIFF
--- a/components/utilities/themeToggle.js
+++ b/components/utilities/themeToggle.js
@@ -8,10 +8,16 @@ const ThemeToggle = () => {
 
   useEffect(() => {
     document.body.dataset.theme = activeTheme;
+    if (activeTheme === "light-mode") {
+      document.documentElement.classList.add("light");
+      document.documentElement.classList.remove("dark");
+    } else {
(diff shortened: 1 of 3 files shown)


imjuangarcia:
@tvst Whenever you have the time, mind checking this PR, and approving if all looks good?

I've been trying out a bunch of different approaches for the CSS refactoring, and some of them rely on Tailwind's `dark` modifier (see https://github.com/streamlit/docs/pull/215 and https://github.com/streamlit/docs/pull/217) so we need to have this in place for those to work as expected.

PR #217, opened 2021-12-24
[Refactor] breadCrumbs CSS

imjuangarcia:
## Description:

This PR is a proof of concept to see whether this approach, using Tailwind directives inside a CSS module with `@apply`, or the one in https://github.com/streamlit/docs/pull/215 are better suited for the Tailwind CSS refactoring and moving away from Sass.

## Prerequisites:

For Tailwind dark mode modifiers to kick in, [the following PR](https://github.com/streamlit/docs/pull/216) needs to be merged.

## The problem with this approach:

None! After [this PR](https://github.com/streamlit/docs/pull/216) is merged, this approach _should_ work, meaning:

* Tailwind styles should work as expected;
* Highly specific styles should be overriden by the Tailwind CSS directives.


tvst (approved):

imjuangarcia:
@tvst Good news! After merging #216, this now works as expected 🎉

One thing to consider though, there's currently [a bug in Tailwind](https://github.com/tailwindlabs/tailwindcss/discussions/2917), where the `dark` modifier doesn't work with `@apply`. It does work inline though, so I moved as many classes as I could to the HTML in 6fa57c9, and kept on the CSS module those that needed more specificity to override the globals. 

LMK if this approach works for you, and I'll get to styling the rest of the components!

PR #218, opened 2021-12-27
Fix on sidebar height

imjuangarcia:
This PR fixes an issue with the sidebar doesn't extending all the way to the bottom of the screen. See image below:

Screen Shot 2021-12-27 at 17 07 18

This PR also fixes an issue on mobile where some elements with higher `z-index` where showing up above the menu. See image below:

Screen Shot 2021-12-27 at 17 00 30

### DIFF
--- a/styles/components/navigation/sideBar.scss
+++ b/styles/components/navigation/sideBar.scss
@@ -5,6 +5,7 @@
   height: calc(100vh - #{$header-height});
   z-index: 0;
   width: 314px;
+  z-index: 2;
 
   * {
     transition: $slow;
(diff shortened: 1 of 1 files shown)


styles/components/navigation/sideBar.scss
```diff
@@ -5,6 +5,7 @@
   height: calc(100vh - #{$header-height});
   z-index: 0;
   width: 314px;
+  z-index: 2;
```
tvst, line 8: LGTM but:

* Line 115 has .sideNav with z-index:11. Should that be here? Not sure how the two interplay...
* At some point we should get all z-index values into a single framework -- or at least bring them all to the same file.

tvst (approved):

styles/components/navigation/sideBar.scss
imjuangarcia: I unified the `z-index` values on c46f93e. I'll keep the `z-index` idea in mind when refactoring the CSS per our convo yesterday!

PR #223, opened 2022-01-04
Make Suggest edits button link to core repo docstring source

snehankekre:
This PR is a very early WIP to make the Suggest edits button on API reference pages link to the original source of the docstring in the core Streamlit repo.

[#71f8d97](https://github.com/streamlit/docs/commit/71f8d9789c53a3632409a2a85c0cb206f001f850) creates an **additional** "Suggest edits" button at the bottom of API reference pages. This is not the desired behavior. Ideally, for API reference pages, we need to override the Helpful component and pass it the GitHub URL instead of `sourcefile`.

@imjuangarcia The Python part of this PR is complete. I could use your input on the JS side of this PR. We need to pass the `functionObject.source` object (containing a Streamlit GitHub URL) from `autofunction.js` to either the Helpful or SuggestEdits component. This should be done only for API reference pages containing docstrings. i.e. pages whose Markdown filename ends with `/library/api/*/*.md`, where * is an alphanumeric string; or equivalently, pages that use the Autofunction component.

### DIFF
--- a/pages/[...slug].js
+++ b/pages/[...slug].js
@@ -2,7 +2,6 @@ import fs from "fs";
 import { join, basename } from "path";
 import sortBy from "lodash/sortBy";
 import React from "react";
-import { useRouter } from "next/router";
 import Link from "next/link";
 import Head from "next/head";
 import { serialize } from "next-mdx-remote/serialize";
@@ -22,6 +21,7 @@ import {
 } from "../lib/api";
 import { getPreviousNextFromMenu } from "../lib/utils.js";
 import useVersion from "../lib/useVersion.js";
+import { useAppContext } from "../context/AppContext";
 import Layout from "../components/layouts/globalTemplate";
 import BreadCrumbs from "../components/utilities/breadCrumbs";
 import SideBar from "../components/navigation/sideBar";
@@ -35,7 +35,6 @@ import ArrowLink from "../components/navigation/arrowLink";
 import Helpful from "../components/utilities/helpful";
 import { H1, H2, H3 } from "../components/blocks/headers";
 import Psa from "../components/utilities/psa";
-import SuggestEdits from "../components/utilities/suggestEdits";
 import FloatingNav from "../components/utilities/floatingNav";
 
 // MDX Components
@@ -72,11 +71,14 @@ export default function Article({
   let versionWarning;
   let currentLink;
   let sourceFile;
+  const autoFunctionSourceFile = useAppContext();
+
   sourceFile =
-    "https://github.com/streamlit/docs/tree/main" +
-    filename.substring(filename.indexOf("/content/"));
+    Object.keys(streamlit).length > 0 && autoFunctionSourceFile
+      ? autoFunctionSourceFile.sourceFile
+      : "https://github.com/streamlit/docs/tree/main" +
+        filename.substring(filename.indexOf("/content/"));
   const maxVersion = versions[versions.length - 1];
-  const router = useRouter();
   const version = useVersion(versionFromStaticLoad, versions, currMenuItem);
 
   const components = {
--- a/pages/[...slug].js
+++ b/pages/[...slug].js
@@ -135,12 +135,22 @@ export default function Article({
 
   if (prevMenuItem) {
     previousArrow = (
-      <ArrowLink link={prevMenuItem.url} type="back" content={prevMenuItem.name} />
+      <ArrowLink
+        link={prevMenuItem.url}
+        type="back"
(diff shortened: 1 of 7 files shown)


imjuangarcia:
@snehankekre What do you think of the approach taken on fef2f29? I'm rendering a new `<Helpful />` utility specific to the `<Autofunction>` component and, when that one is present on the page, I remove the general one we have for all `[...slug]` pages.

snehankekre:
@imjuangarcia This is exactly what I in mind! Do you think we're ready to merge?

snehankekre:
@imjuangarcia Noticed a bug in pages that include content below the `<Autofunction />` component. The Helpful utility is displayed (mid-page) before the rest of the content. See [1](https://deploy-preview-223--streamlit-docs.netlify.app/library/api-reference/widgets/st.button), [2](https://deploy-preview-223--streamlit-docs.netlify.app/library/api-reference/widgets/st.radio), [3](https://deploy-preview-223--streamlit-docs.netlify.app/library/api-reference/widgets/st.slider):

image

imjuangarcia:
> @imjuangarcia Noticed a bug in pages that include content below the `<Autofunction />` component. The Helpful utility is displayed (mid-page) before the rest of the content

@snehankekre Great catch, thank you! The approach I was thinking of won't work, then. I fixed it on 7e41a12 using React Context to store the Autofunction source file globally. I'm looping @tvst in here so he can review the JS piece!

python/generate.py
```diff
@@ -95,6 +95,19 @@ def get_function_docstring_dict(func, funcname, signature_prefix):
                 return_obj['description'] = parse_rst(returns.description) if returns.description else ''
                 return_obj['return_name'] = returns.return_name
                 description['returns'].append(return_obj)
+        
+        description['source'] = []
+        repo_prefix = "https://github.com/streamlit/streamlit/blob/develop/lib"
```
tvst, line 100: @snehankekre : Move into top-level function.

python/generate.py
```diff
@@ -95,6 +95,19 @@ def get_function_docstring_dict(func, funcname, signature_prefix):
...
+
+        try: # For Streamlit commands wrapped by decorator
+            path_parts = func.__wrapped__.__code__.co_filename.partition("/streamlit")
+            line_number = func.__wrapped__.__code__.co_firstlineno
```
tvst, line 104: @snehankekre : May want to do a `while` loop checking whether there's a `.__wrapped__` . In fact, you could even remove the try/except this way:

```
while '__wrapped__' in func:
  func = func.__wrapped__

path_parts = func.__code__.co_filename.partition("/streamlit")
line_number = func.__code__.co_firstlineno
```

python/generate.py
```diff
@@ -95,6 +95,19 @@ def get_function_docstring_dict(func, funcname, signature_prefix):
...
+        except:
+            path_parts = func.__code__.co_filename.partition("/streamlit")
+            line_number = func.__code__.co_firstlineno
+        
+        description['source'] = repo_prefix + path_parts[1] + path_parts[2]+ f'#L{line_number}'
```
tvst, line 109: @snehankekre 

Use f-string:
```
description['source'] = f'{repo_prefix}{path_parts[1]}{path_parts[2]}#L{line_number}'
```

or join:
```
description['source'] = ''.join([
  repo_prefix,
  path_parts[1],
  path_parts[2],
  f'#L{line_number}',
])
```

pages/[...slug].js
```diff
@@ -72,11 +71,14 @@ export default function Article({
   let versionWarning;
   let currentLink;
   let sourceFile;
+  const autoFunctionSourceFile = useAppContext();
```
Comment (tvst) [inline pages/[...slug].js:74]: This could be cleaner:

```
const { sourceFile } = ...
```

tvst (approved):
LGTM after changes.

pages/[...slug].js
Comment (imjuangarcia) [reply pages/[...slug].js]: Fixed on 709f4a9

python/generate.py
snehankekre: Fixed on f73e87b.

python/generate.py
snehankekre: So much cleaner, thank you! I did:

```
while '__wrapped__' in func.__dict__:
  func = func.__wrapped__

path_parts = func.__code__.co_filename.partition("/streamlit")
line = func.__code__.co_firstlineno
```

python/generate.py
snehankekre: I used join on f73e87b.

PR #225, opened 2022-01-04
Debounce resize events

imjuangarcia:
## Description

This PR improves performance of the site, and greatly minimizes (but not fixes entirely) the issue brought by Adrien [here](https://www.notion.so/streamlit/Serious-rendering-problem-with-docs-on-iPhone-03b468ca813c4aa2a01ef83291640f3c). 

The final fix for the issue mentioned above will come in a separate PR as it's more complex and unrelated to this.

### DIFF
--- a/components/layouts/masonry.js
+++ b/components/layouts/masonry.js
@@ -1,3 +1,4 @@
+import { debounce } from "lodash";
 import React, { useState, useEffect } from "react";
 
 const Masonry = ({ children }) => {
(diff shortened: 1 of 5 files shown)


imjuangarcia:
@snehankekre Here's a quick performance win PR for the site, which also helps minimize the flickering issue on iOS Safari.

[Here's a video](https://share.getcloudapp.com/4guZ5X1P) that shows how the site behaves after these changes. As you can see, the issue brought by Adrien is not 100% resolved, but it's greatly reduced. I'll work on a separate PR for the (hopefully) final fix, as it's complex and involves some setup for the code highlighting tool we use (Prism).

snehankekre (approved):
Thank you for working on this fix! I can't speak to the JS, but from the video you shared, the flickering is greatly minimized compared to Adrien's video.

PR #226, opened 2022-01-04
Fix on Header background

imjuangarcia:
This small PR fixes [this issue](https://share.getcloudapp.com/qGurg67r) that happens when you refresh a page that was scrolled:

1. If you scroll, the header becomes sticky, because you've scrolled `20px` from the top of the window;
2. If you refresh the page, those `20px` you scrolled are now gone, causing the layout of the header to break;
3. This PR adds a background to the elements by default, to ensure the header won't overlap with the content underneath it.

### DIFF
--- a/styles/components/navigation/header.scss
+++ b/styles/components/navigation/header.scss
@@ -9,6 +9,7 @@ header {
 
   .container {
     transition: $normal;
+    background: #fff;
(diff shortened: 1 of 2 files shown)


snehankekre (approved):
LGTM!

PR #231, opened 2022-01-06
Fix on code blocks scroll overflow

snehankekre:
This small PR makes the horizontal scrollbar at the bottom of code blocks visible when content overflows on the x-axis.

### DIFF
--- a/styles/components/blocks/code.scss
+++ b/styles/components/blocks/code.scss
@@ -13,8 +13,6 @@
     white-space: pre;
     max-width: 100%;
     overflow: auto;
-    overflow-x: scroll;
-    overflow-x: overlay;
   }
 
   code.undefined {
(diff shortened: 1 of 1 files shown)


styles/components/blocks/code.scss
```diff
@@ -12,7 +12,7 @@
   pre {
     white-space: pre;
     max-width: 100%;
-    overflow: hidden;
+    overflow: auto;
     overflow-x: scroll;
     overflow-x: overlay;
```
imjuangarcia, line 17: Seems like this component has a bunch of `overflow` properties. And, in theory, the `overflow-x` properties below should take precedence over the `overflow: auto;`.

I'd try removing those `overflow-x` properties, and keep only `overflow: auto;` or `overflow-x: auto;` to prevent vertical scrollbars to appear whitin the code block

imjuangarcia (approved):
Approved, but left a comment about repetition. Other than that, looks good!

styles/components/blocks/code.scss
snehankekre: Good catch! I've kept only `overflow: auto;` and have removed the other two `overflow-x` properties.

PR #234, opened 2022-01-10
[Refactor] floatingNav CSS

imjuangarcia:


components/utilities/floatingNav.js
```diff
@@ -112,11 +149,43 @@ const Headings = ({ headings, activeId }) => {
...
+      `}
+      key={`toc-${index}`}
+    >
+      <a
+        href={heading.target}
+        className={`
+          border-b-0
+          inline-block truncate w-40
+          leading-4
+          hover:opacity-70 hover:border-b-0 hover:no-underline
```
tvst, line 183: Move the above into a class

components/utilities/floatingNav.js
```diff
@@ -112,11 +149,43 @@ const Headings = ({ headings, activeId }) => {
...
+        ${
+          heading.level === "H1" || heading.level === "H2"
+            ? "pl-6"
+            : heading.level === "H3"
+            ? "pl-8"
+            : heading.level === "H4"
+            ? "pl-10"
+            : heading.level === "H5"
+            ? "pl-12"
+            : "pl-14"
```
tvst, line 172: Move the above into a class.


For the heading picker, you could go different routes:

heading.level == "H1" ? styles.headingH1 : etc

styles[heading.level]

use "heading1" as a style
:global(heading1)

tvst (approved):
LGTM after changes

components/utilities/floatingNav.js
imjuangarcia: Fixed on c8573ad

components/utilities/floatingNav.js
imjuangarcia: Fixed on c8573ad

PR #235, opened 2022-01-10
[Refactor] GDPR CSS

imjuangarcia:


components/utilities/gdpr.js
```diff
@@ -68,15 +70,65 @@ const GDPRBanner = (gdprData) => {
...
+            className="
+              bg-white
+              border-0
+              rounded-md
+              px-4 py-3
+              text-red-70 text-sm font-bold tracking-normal
+              block
+              cursor-pointer
+              mb-0
+              hover:opacity-90 hover:scale-105
```
tvst, line 127: Please move all styles into the css file

tvst (approved):

components/utilities/gdpr.js
imjuangarcia: Fixed on 6591c2a

PR #243, opened 2022-01-11
Revert "Update Cloud docs for new sign-in method"

randyzwitch:
Reverts streamlit/docs#228


randyzwitch:
Technically, this goes out today, so closing instead of reverting. Will check calendar next time :(

PR #247, opened 2022-01-12
[Refactor] socialCallout CSS

imjuangarcia:
Changed the background colors for the icons a bit, so we could use Streamlit's color palette (except for Instagram). Is that ok, or do you want me to use custom colors for all (like we do in prod)?


tvst:
>  Is that ok, or do you want me to use custom colors for all (like we do in prod)?

I think it looks better this way!

tvst (approved):

PR #249, opened 2022-01-12
[Refactor] Search CSS

imjuangarcia:
Had to create a specific `searchModal.css` file, and import it globally, so we can style the classes Algolia uses internally.


components/utilities/search.js
```diff
@@ -150,20 +152,26 @@ const Search = () => {
 
     return (
       <article
-        className={`item ${
-          props.hit.__position === indexFocus ? "focused" : ""
-        }`}
+        className={`
+          ${styles.HitContainer}
+          ${props.hit.__position === indexFocus ? styles.ActiveHit : ""}
+        `}
```
tvst, line 158: Use `classNames()` utility.

Same below.

components/utilities/search.js
```diff
@@ -196,15 +204,23 @@ const Search = () => {
...
               // left: '-40em'
             }}
-            className="algolia"
+            className={styles.ModalContainer}
           >
             <FocusTrap>
               <div
-                className="modalContainer"
+                className={styles.Modal}
+                style={{ background: "rgb(163, 168, 184, .6)" }}
```
tvst, line 212: Why use inline styles? Can this be put in a class instead? Same below.

Also: why not use tailwind? For example: `bg-gray-40/60`

tvst:
This breaks some of the existing design.

### Text input background color

Today:
image

This PR:
image

---

### Underline and result category styling

Today:
image

This PR:
image

---

tvst (changes requested):

components/utilities/search.js
imjuangarcia: Fixed on f0b5331

imjuangarcia:
> This breaks some of the existing design

Great catch, thanks! Both were fixed on 57e8eef and 226ac21

components/utilities/search.js
```diff
@@ -150,20 +153,29 @@ const Search = () => {
...
-          <section className="image_container bg-yellow-90">
-            <div className={`icon-${icon}`}>
-              <i>{icon}</i>
+        <a
+          className={classNames("not-link", styles.HitLink)}
+          href={props.hit.url}
+        >
+          <section className={styles.IconContainer}>
+            <div
+              className={`
```
tvst, line 168: nit: use `classNames`

components/utilities/search.js
```diff
@@ -231,11 +250,20 @@ const Search = () => {
   }
 
   let searchBar = (
-    <section className="block-search">
-      <section className="hot-key small" onClick={searchClicked}>
-        <i>search</i>
-        <p className="search-text">Search</p>
-        {windowWidth > 1024 && <p className="command">{hotkey}</p>}
+    <section
+      className={`
```
tvst, line 254: `classNames`

tvst (approved):

components/utilities/search.js
imjuangarcia: Fixed on bd22ce4

components/utilities/search.js
imjuangarcia: Fixed on bd22ce4

PR #250, opened 2022-01-12
Add paragraph about cache clearing

jrieke:
Most of the content comes from [the release blog post](https://www.notion.so/streamlit/What-s-new-in-Streamlit-254a2095c53b4745b24f62a85884b72d). Feel free to adapt as you like!


snehankekre (approved):
LGTM @jrieke! The only change I've made is to replace `foo` with `square`. I'll merge this PR after the `1.4.0` release today.

PR #252, opened 2022-01-13
Fix singleton and memo docstrings in v1.4.0

snehankekre:
[As of Streamlit v1.4.0](https://github.com/streamlit/streamlit/pull/4263),  `@st.experimental_memo` and `@st.experimental_singleton` are now implemented as  `__call__` functions within singleton MemoAPI and SingletonAPI objects, respectively, rather than as standalone functions. 

This PR makes `generate.py` pull docstrings for memo and singleton from  their respective `__call__` functions.  If it were not for this PR, we would see a warning callout that reads _This method did not exist in version 1.4.0 of Streamlit._.


randyzwitch (approved):
Looks reasonable to me

PR #260, opened 2022-01-14
Fix on reference card margins

imjuangarcia:
This PR fixes an issue reported by Johannes, where the padding for cards without images was too small

Example:

image


snehankekre (approved):
LGTM!

PR #261, opened 2022-01-16
Update api-cheat-sheet.md

EswarDivi:
Added key argument  in st.form


snehankekre (approved):
Thank you, @EswarDivi. Your contribution makes it clear that each form must have its own key.

PR #262, opened 2022-01-18
Add memo.clear and singleton.clear to API reference

snehankekre:
### Description
Add memo.clear and singleton.clear to API reference.

Streamlit 1.4.0 implements two new functions `st.memo.clear()` and `st.singleton.clear()` , which clear the caches for all functions with memo and singleton, respectively. [MADE has requested](https://www.notion.so/streamlit/Make-st-memo-clear-show-up-in-API-reference-c0f25db960da42d598e3ed295aa51e30) that we include them in the API reference.

### Changes
- Modifies `generate.py` to include docstrings for the `clear()` method. We usually get docstrings of `st.command`. This PR makes an exception for `st.command.clear`.
- Creates cards for both functions on the API reference landing page.
- Creates API reference pages for both functions.


randyzwitch (approved):
Looks good, thanks! Nice catch on the spelling mistake as well :)

PR #263, opened 2022-01-18
Update 'No dependency file' Section

raethlein:


randyzwitch:
This is a product decision for sure, but it feels weird for us to document behavior from indeterminate states. Meaning, this feels like an internal design decision that could just as easily be changed in the future; should we just be telling users to explicitly set the version in their requirements or risk getting undesired behavior?

raethlein:
@randyzwitch good point, that would also be totally fine for me and I don't have a strong opinion here. I just wanted to create this PR so it does not get lost that the Section seems to need an update

atoader (approved):

snehankekre:
@raethlein @randyzwitch I've added a small paragraph on 39aa3bb to clarify that using a dependency file and pinning a version of Streamlit is the best practice, if users want to avoid undesired behavior. WDYT?

snehankekre (approved):

randyzwitch:
Looks good, thanks @snehankekre!

PR #272, opened 2022-01-21
[Refactor] globalTemplate deletion

imjuangarcia:


tvst (changes requested):
I can't tell why just from looking at the code, but this PR adds a big top-margin above breadcrumbs

### This PR
image


### Prod
image

tvst (approved):
Actually LGTM because I think this issues came from another PR... not sure which. 

Can you look into them?

imjuangarcia:
> Actually LGTM because I think this issues came from another PR... not sure which.
> 
> Can you look into them?

@tvst The issue is that this PR doesn't have the latest changes on `main`, so it's missing some updates, like the header refactoring, where I removed some extra margin bottom.

**This PR, without the latest additions to `main`:**
Screen Shot 2022-02-10 at 19 05 17

**Prod, with the refactored header:**
Screen Shot 2022-02-10 at 19 06 01

PR #274, opened 2022-01-21
[Refactor] Masonry CSS

imjuangarcia:


components/layouts/masonry.js
```diff
@@ -37,7 +39,10 @@ const Masonry = ({ children }) => {
   }, []);
 
   return (
-    <section className="masonry" style={{ "--max-height": height + "px" }}>
+    <section
+      className={styles.Container}
+      style={{ "--max-height": height + "px" }}
```
tvst, line 44: Is this CSS variable actually used anywhere? I did a quick search on Github but couldn't find it

tailwind.config.js
```diff
@@ -8,16 +8,6 @@ module.exports = {
...
   theme: {
-    screens: {
-      // Remove largest breakpoint, while keeping everything else at Tailwind defaults.
-      // https://tailwindcss.com/docs/breakpoints
-      sm: "640px",
-      md: "768px",
-      lg: "1024px",
-      xl: "1280px",
-      //'2xl': '1536px',
-    },
```
tvst, line 19: Is this on purpose?

tvst (approved):
Approved but PTAL at comments

components/layouts/masonry.js
imjuangarcia: I don't think it is... I deleted it on [64a7579](https://github.com/streamlit/docs/pull/274/commits/64a7579bb583bbaf2138605ec2e881c00703ac7b)

tailwind.config.js
imjuangarcia: Yes. I deleted it because we were using all Tailwind defaults, but only removing the `2xl` breakpoint for larger screens. I added it back so we can use a three-column design here on large screens

PR #276, opened 2022-01-21
[Refactor] Row deletion

imjuangarcia:


components/utilities/breadCrumbs.module.css
```diff
@@ -14,6 +14,10 @@
...
 
 .Separator {
-  @apply dark:text-gray-70 ml-2;
+  @apply ml-2;
   @apply text-gray-50 !important;
 }
+
+:global(.dark) .Separator {
+  @apply text-gray-70 !important;
+}
```
tvst, line 23: Is this part of this PR on purpose?

tvst (approved):

components/utilities/breadCrumbs.module.css
imjuangarcia: It is! When we refactored the `breadCrumb` component, I didn't realize the `/` separator changed its color slightly on dark mode, so I snuck it in here 😄

PR #279, opened 2022-01-24
Add camera_input to cheatsheet

tvst:


snehankekre (approved):
I should have fixed it in #269, but didn't. Thanks!

PR #285, opened 2022-01-27
Fix on sharing image URL

imjuangarcia:
This PR intends to fix the issue brought by @CharlyWargnier where thumbnail images aren't showing up when sharing content on twitter

### DIFF
--- a/pages/404.js
+++ b/pages/404.js
@@ -32,11 +32,11 @@ export default function Home({ window, menu }) {
         <meta content="summary_large_image" name="twitter:card" />
         <meta
           property="og:image"
-          content={`${process.env.NEXT_PUBLIC_HOSTNAME}/sharing-image-facebook.jpg`}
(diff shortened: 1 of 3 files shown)


snehankekre (approved):
LGTM @imjuangarcia! I had assumed, incorrectly, that we would have to add unique thumbnails to each article page.

imjuangarcia:
> LGTM @imjuangarcia! I had assumed, incorrectly, that we would have to add unique thumbnails to each article page.

That'd be another way to go about it. We have the ability to do that on the website, and default to the logo image when nothing is added. For the docs' site, I feel it's going to be _a lot_ of work, that's why I kept the default image on all pages, but LMK if you prefer otherwise!

snehankekre:
In that case, let's not pursue it. There are other eng and content tasks that take precedence.

randyzwitch:
For sure on the custom images part, maybe we can get an intern to do that

PR #287, opened 2022-01-28
[Refactor] iconHeader && Important && Note && NoteSplit && Tip && Warning CSS refactor

imjuangarcia:
This PR refactors a bunch of components that are fairly similar amongst them, and removes an almost duplicate `note` component.


tvst:
These icons aren't quite centered:

image

Here's the original, for reference:

image

tvst (approved):
The line height in these code boxes is larger in the PR than today. (2em vs 1.75em)

Do the docs use 1.75em as the line-height in general? If so, we should change the Tailwind config to make regular text have 1.75 height

imjuangarcia:
> These icons aren't quite centered

Fixed on [ad2020c](https://github.com/streamlit/docs/pull/287/commits/ad2020c0e957b3dd29e9368d50ae1b925f30f038)

> The line height in these code boxes is larger in the PR than today. (2em vs 1.75em)

Fixed on [f961d47](https://github.com/streamlit/docs/pull/287/commits/f961d478648e75d927367472acb48d08ffce434b)

> Do the docs use 1.75em as the line-height in general? If so, we should change the Tailwind config to make regular text have 1.75 height.

I agree! I think it'd be good to make changes like these, that affect multiple components, when we remove the global styles, with all the PRs merged, to ensure we don't accidentally break anything. I'll try it out after I merge everything!

PR #294, opened 2022-02-01
Add a note to public-component to suggest github tag

blackary:
I'm not 100% sure this is the best place for this, but this adds a suggestion to add `#streamlit-component` as a tag to any new components, so that they will all show up [here](https://github.com/topics/streamlit-component).


randyzwitch:
I'll defer to @snehankekre here, but would it make sense to add a picture of the tag, just to make sure people understand where to add it?

Approving, I'm fine either way

randyzwitch (approved):

snehankekre:
Thanks @blackary and @randyzwitch. I've added a GIF that walks through the steps of adding the tag to one's component repo.

snehankekre (approved):
Thanks @blackary and @randyzwitch. I've added a GIF that walks through the steps of adding the tag to one's component repo.

PR #295, opened 2022-02-01
Fix on indentation for navbar

imjuangarcia:
This PR fixes an issue where `<h4>` tags on the `floatingNav` component aren't indenting properly.

**Before:**
Screen Shot 2022-02-01 at 17 05 44

**After:**
Screen Shot 2022-02-01 at 17 08 25


snehankekre (approved):
Thank you for the fix! :pray:

PR #302, opened 2022-02-07
Add docs for new resource limits error page

snehankekre:
### :books: Context
MADE [launched a better](https://www.notion.so/streamlit/Better-Argh-out-of-resources-error-page-445f221990974e82b52ae529a77ed529) "Argh" out-of-resources error page and we need to [document](https://www.notion.so/streamlit/Document-new-resource-limit-error-page-50e65938a9c243f4907a30cca839985b) it.

### :brain: Description of Changes
- Created a separate section for "Resource limits" and documented the various pages seen depending on who (dev VS viewer) is viewing the app and what version of the Streamlit (`<1.1.0` VS `>=1.1.0`) the app uses.
- Updated the relevant troubleshooting section and KB article.


randyzwitch (approved):

jrieke (approved):
LGTM!

(Random thinking for the future, not for this project: Wondering if it's better to document features like this with screenshots or not. Sure, it's nice to show to the user but it also means we need to update the docs each time we change something. Maybe it is better to just say "we show you an error page". Anyway, just some thoughts, don't even know what I like better myself... Moving on :running_man:)

randyzwitch:
I think screenshots are desirable, since people tend to scan rather than read

snehankekre:
> Sure, it's nice to show to the user but it also means we need to update the docs each time we change something. Maybe it is better to just say "we show you an error page".

While I agree with this sentiment and wish the screenshots didn't have to be updated so often, our users and customers have made it clear that screenshots are desirable. We've updated pages to include screenshots in the recent past at the express request of customers.

PR #313, opened 2022-02-15
Create Codespace config

snehankekre:
This PR creates the necessary configuration files to define a Codespaces environment for this repo. It sets up VSCode, Python, Node.js, other dependencies, and spins up a docs development server with a browser preview.

I leave it up to the consensus of reviewers to decide if the costs of using Codespaces outweigh its benefits and, consequently, if this PR should be merged.


snehankekre:
Going to merge this as it is an easier alternative for internal contributors who find it tedious or time consuming to set up a local dev env. We're not charged if we don't use it. Having it as an option for other internal contributors is beneficial.

Post-merge:
To launch a codespace, click on Code -> New  codespace -> Create codespace

image
image

PR #321, opened 2022-02-24
Fix on newsEntry background color

imjuangarcia:
This PR fixes [this issue](https://www.notion.so/streamlit/Revert-NewsEntry-utlity-color-1fd0e32eb4574b978ecb7cbf97914a7d) mentioned by @snehankekre, where tiles with `unset` background color would default to the light blue color


snehankekre (approved):
LGTM! Thanks @imjuangarcia :smile:

PR #328, opened 2022-02-28
Fix issue with Release Candidate URLs on docs page

snehankekre:
This PR fixes an [issue](https://www.notion.so/streamlit/Issue-with-Release-Candidate-URLs-on-docs-page-7ce491a6382b40a582a4787f863c02ce) with Release Candidate URLs on docs page noticed by @imjuangarcia.

Release Candidate URLs don’t seem to work as expected on the autofunction component (See video of the bug [here](https://share.getcloudapp.com/E0uowQ6K)). These URLs seem to be generated differently than other versioned URLs, because [they add the number on the URL](https://share.getcloudapp.com/KouYg2yG), which also throws an error on the build because these numbered pages aren’t (and shouldn’t be) on `menu.md`.

It seems like the RCs were generated erroneously while trying to create the 1.6.0 release. They shouldn't be included in the library selector dropdown. We use [build.py](https://github.com/streamlit/docs/blob/main/python/build.py)
 to pick out only major releases, and don't include minor releases (patches) and RCs.

As an immediate solution, this PR deletes all docstrings upto v1.5.0, and runs `python generate.py 1.6.0` instead of `make docstrings`.

TODO: A permanent fix is to patch `build.py` to ignore release candidates.


imjuangarcia (approved):
LGTM! Thanks a lot for the quick turnaround 🙏🏻

PR #333, opened 2022-03-03
Fix on z-index for GDPR banner

imjuangarcia:
This PR fixes an issue reported by Anna Kostenko, where the sidebar would cover the GDPR banner, not allow the user to decline it.

**Before:**
Selection_248
Selection_249

**After:**
Screen Shot 2022-03-03 at 13 28 55


snehankekre (approved):
LGTM!

PR #339, opened 2022-03-12
Update code example to resolve NameError

pedantic-curmudgeon:
Previously this example would raise the following:
```
NameError: name 'st' is not defined
```


snehankekre (approved):
Thank you for your contribution and attention to detail, @pedantic-curmudgeon! :slightly_smiling_face:

PR #345, opened 2022-03-24
switch to dot notation to avoid error in caching with database example doc

discdiver:
['postgres'] notation resulted in an error that reads in part:

```
While caching the body of init_connection(), Streamlit encountered an object of type builtins.weakref, which it does not know how to hash.
```


snehankekre (approved):
Thank you for spotting and fixing the bug, @discdiver! I've verified that using the dot notation fixes the error. I'll update the notation in our other tutorials too.

Issue #363, opened 2022-04-02
Documentation text color issue


osbm:
Here is the [link](https://docs.streamlit.io/library/api-reference/layout/st.columns)

This bit in the documentation when using the black theme is not readable.

![image](https://user-images.githubusercontent.com/74963545/161400282-59fb0028-40ce-48e7-b09c-ad837c64f92d.png)

snehankekre:
Thank you for reporting this bug, @osbm! We'll work on a fix shortly.

PR #364, opened 2022-04-04
Fix color on dark mode for table content

imjuangarcia:
This PR fixes the issue mentioned [here](https://github.com/streamlit/docs/issues/363) about text legibility on dark mode within tables


snehankekre (approved):
Wow, thank you for fixing this so quickly! LGTM to merge :smile:

PR #365, opened 2022-04-05
Style Note and Warning callouts in API reference

snehankekre:
This PR fixes two issues ([1](https://www.notion.so/streamlit/st-set_page_config-note-box-not-rendered-correctly-8de759688e3b40aab3d98f156bb4f7fd), [2](https://www.notion.so/streamlit/st-columns-API-ref-38c5d360fa134702b1b98ffc3c932570)) whereby callouts such as `Note` and `Warning` in API reference docstrings are rendered as paragraph text without any CSS styling like their React component counterparts.

The issues arises because the content in the docstring is not pulled from markdown files (so Tailwind styling does not apply). It is instead pulled from `python/streamlit.json`, that in-turn is populated by `python/generate.py` which parses the `.. note:: ` and `.. warning::` admonitions from Python scripts in the Streamlit library repo. E.g. [st.set_page_config](https://github.com/streamlit/streamlit/blob/38256624c59a41123eebab98ddcfc176daf3cd3a/lib/streamlit/commands/page_config.py#L40-L42).

## Before
image
image

## After
image
image

### DIFF
--- a/styles/admonition.scss
+++ b/styles/admonition.scss
@@ -1,0 +1,32 @@
+// Styles for notes, warnings, etc., inside API reference page
+div.admonition {
+  @apply p-4 rounded-md mt-8 mb-12;
+
+  // Title
+  p.first.admonition-title {
+    @apply text-lg font-bold;
+  }
+
+  // Text
+  p {
+    @apply mb-0;
+  }
+
+  // Note
+  &.note {
+    @apply bg-lightBlue-10 dark:bg-lightBlue-100/30;
+
+    p.first.admonition-title {
+      @apply text-lightBlue-80 dark:text-lightBlue-60;
+    }
+  }
+
+  // Warning
+  &.warning {
+    @apply bg-orange-10 dark:bg-orange-100/30;
+
+    p.first.admonition-title {
+      @apply text-orange-70 dark:text-orange-70;
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


imjuangarcia:
@snehankekre Thanks for fixing this! Would it be possible to also add the icons we use for these `<Note />`, `<Warning />`, `<Tip />`, etc., components? Examples below 👇

Screen Shot 2022-04-05 at 08 29 39

Screen Shot 2022-04-05 at 08 29 57

snehankekre:
@imjuangarcia I'm unsure how to add the icons via `styles/text.scss`. 😅  Would appreciate any pointers or if you could take a stab at it..

imjuangarcia:
@snehankekre I see! Since we aren't rendering React components for these, it won't be possible to add the icons.

I just pushed a commit with a bit of refactoring on the CSS to avoid repetition, and also updated paddings a bit to accommodate for the lack of the icon. Let me know what you think, and feel free to merge if all looks good!

imjuangarcia (approved):

snehankekre:
> Since we aren't rendering React components for these, it won't be possible to add the icons.

That's what I suspected too. Thanks for improving the styling @imjuangarcia! I'm fine with merging these changes if Randy approves as well. Not having the icons on few API reference pages is not an issue IMO.

randyzwitch (approved):

randyzwitch:
Looks good to me, I agree that the icons are not tremendously important in this context

PR #385, opened 2022-05-02
Specify location for packages.txt

stefsmeets:
This edit specifies the location of `packages.txt`.

From the documentation it was not clear to me that `packages.txt` must be in the root directory of the repository. Especially because `requirements.txt` will be read from a subdirectory.


snehankekre (approved):
Thanks for the clarifying PR, @stefsmeets! I've verified that `packages.txt` has to be in the root dir of the repo to be installed, while `requirements.txt` can be installed from sub dirs.

The former is news to me, so I'm checking with the team if it is intentional or a bug.

PR #396, opened 2022-05-27
Offline docs in PDF version

imjuangarcia:
## 📚 Context

This PR introduces a couple of scripts to export our docs into PDF format. It's something that [had been requested in the forum](https://discuss.streamlit.io/t/streamlit-official-documentation-in-pdf-format/25342/3) before. Steps to use locally:

```
git fetch
git checkout docs-pdf-conversion
npm install

// With npm run dev running in a different tab to have the local server up
npm run offline-docs
```

This is still a WIP due to a variety of reasons:

  1. Due to the large quantity of PDFs to generate and each page's weight, Puppeteer will "randomly" export some of them without the styling;
  2. Also per the above 👆 we can't add these scripts while the site is built, because it'll often fail if a PDF isn't generated properly;
  3. Currently, the PDFs get regenerated every time we run the script.

There are several improvements to be made, such as:

* [x] Add a `media: print` stylesheet to style the pages consistently (added on https://github.com/streamlit/docs/pull/407);
* [x] ~~Adding the scripts that generate the PDFs on a `pre-commit` or `pre-push` hook;~~ 🔴 No need for now as it'll be locally generated.
* [x] ~~Check the `<lastmod>` entry in `sitemap.xml` to determine if the page has changed before regenerating the PDF;~~ 🔴 No need for now as it'll be locally generated.
* [x] ~~Reducing the size of the final document;~~ 🔴 No need for now as it'll be locally generated.
* [x] Order the pages the same way as they appear on docs.streamlit.io

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small 

## 🌐 References

* https://discuss.streamlit.io/t/streamlit-official-documentation-in-pdf-format/25342/3


imjuangarcia:
@snehankekre Made some updates to this PR. Running the command locally _should_ work now, without timing out, and also the pages should be styled correctly now that we've merged the `print` stylesheet.

A couple things to note:

1. For the PDF generation to work properly and be styled consistently, I had to replace the production url (`https://docs.streamlit.io`) with the `localhost` URL. That means we can't add it as a post-deploy script in Netlify because we're not running the local server there;
2. Since Netlify might not be an option ☝🏻, I added the command to generate the PDF on a `pre-push` hook. This means your changes aren't pushed to the branch until the script is completed (currently takes ~1-2 minutes). Is that desirable? Or do we not want to enforce this delay?
3. The `pre-push` hook doesn't add, commit and push the file to the repo (for now) because it's a rather big (~80mb) file. Is still the plan to host it on the repo, given its size? Or should we upload it to S3?

snehankekre:
@imjuangarcia Awesome!

2. We probably don't want to enforce this delay for every push. Could we make it so that only certain contributors can run the pdf generation on specific PRs? For example only for PRs related to a new core release
3. Given its size, we should not host it on the repo. S3 might make more sense. WDYT @randyzwitch?

> Running the command locally should work now, without timing out

I fetched the latest changes and encountered a `TimeoutError` after running for 3-4 mins:

```
Done! Created PDF for libgl
Done! Created PDF for no-matching-distribution
/Users/skekre/docs/node_modules/puppeteer/lib/cjs/puppeteer/common/helper.js:232
    const timeoutError = new Errors_js_1.TimeoutError(`waiting for ${taskName} failed: timeout ${timeout}ms exceeded`);
                         ^
TimeoutError: waiting for Page.printToPDF failed: timeout 30000ms exceeded
    at Object.waitWithTimeout (/Users/skekre/docs/node_modules/puppeteer/lib/cjs/puppeteer/common/helper.js:232:26)
    at Page.createPDFStream (/Users/skekre/docs/node_modules/puppeteer/lib/cjs/puppeteer/common/Page.js:2143:49)
    at Page.pdf (/Users/skekre/docs/node_modules/puppeteer/lib/cjs/puppeteer/common/Page.js:2156:37)
    at getPDFs (/Users/skekre/docs/scripts/build-offline-docs.js:88:14)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Promise.all (index 42)
```

Maybe lines 88-91 of `build-offline-docs.js` should read:

```js
  await page.pdf({
    path: `public/pdf/pages/${url.slug}.pdf`,
    format: "a4",
    timeout: 0,
  });
```

snehankekre:
Ok what worked was deleting my local branch, fetching the remote branch, disabling timeout, and running the script. It took around 10 mins to generate the final pdf -- not too bad! 

Another improvement could be to sort pages in the order in which they appear in the docs (following the `Next:` arrow link at the bottom of pages). The pages in the final pdf appear to be randomly sorted

randyzwitch:
For the time being, I would think about just writing proper instructions for people to use to do this themselves, up to possibly creating a Docker image for ease-of-generation. We can then think about how to do this as a separate process that doesn't hold up the web development piece, creates an archive, etc.

imjuangarcia:
@snehankekre 

> Ok what worked was deleting my local branch, fetching the remote branch, disabling timeout, and running the script.

Ok, so then there's no need to add the `timeout: 0`, right? Either way it's working for me.

> Another improvement could be to sort pages in the order in which they appear in the docs (following the Next: arrow link at the bottom of pages). The pages in the final pdf appear to be randomly sorted

This is rather complex, because I'm not using puppeteer anymore to navigate each page, and instead I'm taking the URLs from [our sitemap](https://docs.streamlit.io/sitemap.xml). What I just did on [cbb3cac](https://github.com/streamlit/docs/pull/396/commits/cbb3cac4629dd8278e70d6b5c809c85761243af3) is modifying the `slug` parameter to have indexes at the beginning, so it's easier to group them on each category (`Library`, `KB` and `Streamlit Cloud`). Would that work, or is it too fragile/arbitrary? If that's not good enough let me know and I'll try to find a workaround for it!

snehankekre:
@imjuangarcia 
> Ok, so then there's no need to add the timeout: 0, right? Either way it's working for me.

I guess not! It might have been due to a bug in my environment.

> Would that work, or is it too fragile/arbitrary? If that's not good enough let me know and I'll try to find a workaround for it!

It would definitely work as a first iteration. We could improve the ordering of pages in a later PR. 

Since there's almost a 1:1 mapping between the urls in `content/menu.md` and `public/sitemap.xml`, is possible within the script to fetch the links from the sitemap, sort them by the order in which they appear in `content/menu.md`, and then generate PDFs?

Unrelated: I noticed the sitemap wasn't updated since Oct, 2021. So I updated it locally and pushed the changes on a951d85.

imjuangarcia:
@snehankekre Using `menu.md` is a good idea! I found a way to improve the sorting on [94e693a](https://github.com/streamlit/docs/pull/396/commits/94e693a7ebd120f0e485c747b12256402b0544cd) so it works as you suggested. Let me know what you think!

Also, tried to adjust the command we use to generate the sitemap automatically, `next-sitemap`, to see why it wasn't updating since October, but I'd rather do that in a separate PR to avoid making this one more complex than it is.

snehankekre:
@imjuangarcia The pdf generation works wonderfully now! 🎉 

> Were you referring to the file itself, or the URLs inside the sitemap?
Both. The file metadata indicated it hadn't been updated since Oct, 2021. URLs of pages that were created after Oct, 2021, weren't included the sitemap until this PR (e.g. https://docs.streamlit.io/library/get-started/multipage-apps, https://docs.streamlit.io/library/api-reference/personalization/st.experimental_user)

imjuangarcia:
@snehankekre Ok, so then I guess we could merge this? Or are there any other adjustments you want me to make to it?

snehankekre (approved):
@imjuangarcia Yeah, feel free to merge this PR when ready 😃 The sitemap issue can be tackled as a separate PR. A `pre-push` hook could work. It only takes a couple of seconds to generate the sitemap.

PR #406, opened 2022-06-03
Fix on versioned URLs starting on 1.10

imjuangarcia:
This PR fixes [this issue](https://www.notion.so/streamlit/API-version-selector-not-showing-latest-version-by-default-ef6b19e9e21c46ca8064ac1fe06e22c4) reported by @snehankekre

### DIFF
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -106,7 +106,6 @@ const Autofunction = ({
 
   if (streamlitFunction in streamlit) {
     functionObject = streamlit[streamlitFunction];
-    const sourceFile = useSourceFile(functionObject.source);
 
     if (
       functionObject.description !== undefined &&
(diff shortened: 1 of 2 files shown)


snehankekre:
@imjuangarcia I think we need to add the following back to line 109 of `autofunction.js` so that users can click on Suggest edits and be directed to the source code in the Core repo:

```js
const sourceFile = useSourceFile(functionObject.source);
```

imjuangarcia:
> @imjuangarcia I think we need to add the following back to line 109 of `autofunction.js` so that users can click on Suggest edits and be directed to the source code in the Core repo rather than a Markdown file in the docs repo:
> 
> ```js
> const sourceFile = useSourceFile(functionObject.source);
> ```

You're right! Thanks for adding that back 🙏🏻

PR #431, opened 2022-07-13
Better eu check

sfc-gh-tteixeira:
## 📚 Context

We currently check if user is in Europe (for GDPR banner) based on timezone, by looking at the TZ offset. This is a bit hacky, brittle, and hard to read. This PR changes the code to look at the timezone name instead, which, for Europe is always `Europe/something`.
 
## 🧠 Description of Changes

Change "is in Europe" check to use timezone name.

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

- [Idea copied from this Gist](https://gist.github.com/Cosbgn/f18af53f223092f659818bd565194825?permalink_comment_id=3079849#gistcomment-3079849)


imjuangarcia (approved):
Way cleaner! Love it. Tested locally and the behavior seems unaffected 👍🏻

PR #441, opened 2022-07-25
Correct description of HSTS

sndyuk:
## 📚 Context

This is a trivial fix.


snehankekre (approved):
Good catch, @sndyuk! Thank you for the fix 😃

PR #468, opened 2022-09-06
Add a new `Collapse` UI component

snehankekre:
## 🧠 Description of Changes

- Adds a new `Collapse` UI component

image

image

@imjuangarcia could use your help with improving the styling ✨ 

### Usage

```md
<Collapse title="your title">

... your content

</Collapse>
```

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small


snehankekre:
@imjuangarcia We need to make it so that anchor links for headings work. Currently, the following doesn't work:

```md
<Collapse title="Install Streamlit on macOS/Linux">

## Install Streamlit on macOS/Linux

</Collapse>
```

The anchor link for `Install Streamlit on macOS/Linux` isn't created.

imjuangarcia:
@snehankekre FYI: I'm hoping to discuss this with @tvst tomorrow (along with the hover effect in `refCard` components), mostly to see whether we'd like this expander to look more like our existing `st.expander` widget, or more like Notion's toggle, and then I'll tackle the UI for it! Do you have any preference regarding how it should look like?

snehankekre:
Thanks for the heads up! The motivation to create the component was to avoid manually using something like:

<summary><details>
This is some text
</details>
</summary>

```markdown
<summary><details>
This is some text
</details>
</summary>
```

which is used in a few [places](https://docs.streamlit.io/library/get-started/main-concepts#pages). When rendered it looked like Notion's toggle. Hence the similar UI in the current version of this component.

I don't have a preference either way, but would like the UI and look to be consistent with that of the rest of the docs.

imjuangarcia:
Oh, I see! Then it might make more sense to make it look a bit more like our `st.expander` widget. I'll do that, then, and let you know when it's ready!

components/blocks/collapse.module.css
```diff
@@ -0,0 +1,40 @@
...
+  @apply border border-gray-30 rounded-md p-3;
+}
+
+.PanelTitle {
+  @apply list-none flex items-center mb-0 transition ease-in-out duration-75 hover:opacity-80 cursor-pointer;
+}
+
+.Title {
+  @apply flex-1;
+}
```
snehankekre, line 15: - When the `<Collapse>` element is expanded, the delineation between the title and rest of the content isn't clear IMO because there isn't a horizontal rule separating them nor a difference in font size / color: 
    image
- What are you thoughts on slightly increasing the title font size:
```css
.Title {
  @apply flex-1 text-lg;
}
```

image

snehankekre:
@imjuangarcia The new look is awesome! 😎 

- Thank you for breaking down the issues with having titles inside `<Collapse>` elements.  Let's not use this component with headings inside them 👍 

- On 21c40c1, I added an example usage of the component to the Installation page -- if you want to play with it.

- With the current implementation, the cursor pointer showed up everywhere within the container when expanded. On 05c47aa, we show the cursor pointer only on the title.

- I found it hard to tell there was a border around the `PanelHeading` with `border-gray-20`, so I darkened it slightly with `border-gray-30` on cc58d67. WDYT? Although, it deviates from `st.expander` style, it makes the border perceptible IMO:
   - Before: image
   - After: image

components/blocks/collapse.module.css
imjuangarcia: I like it!

imjuangarcia (approved):
Looks great, love these last updates you did to the font size and border color! 👍🏻

PR #470, opened 2022-09-06
Add optional target attribute to some components

snehankekre:
## 📚 Context

The `Read More ->` links, `View all updates` button, and `View forum` button on the docs homepage open the links in the same tab. There's a small but non-zero risk of users opening the links and not returning to the docs after.

This PR updates `button`, `newsEntry`, `noteSplit`, and `arrowLink` components to include an optional `target` attribute. When not specified, `target = "_self"`.  This change allows us to set `target="_blank"` where required (e.g. in `pages/index.js`), so that the links open in a new tab.

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small


components/blocks/button.js
```diff
@@ -2,12 +2,20 @@ import Link from "next/link";
...
-const Button = ({ children, link }) => {
-  return (
-    <Link href={link}>
-      <button className={styles.Button}>{children}</button>
-    </Link>
-  );
+const Button = ({ children, link, target }) => {
+  if (target) {
+    return (
+      <a href={link} target={target}>
```
imjuangarcia, line 8: Should we add `rel="noopener noreferrer"` when `target` is set to `_blank`? It was a best practice a while back to prevent cross-scripting, but I believe now [browsers handle it automatically](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener)?

components/blocks/button.js
```diff
@@ -2,12 +2,20 @@ import Link from "next/link";
...
 import styles from "./button.module.css";
 
-const Button = ({ children, link }) => {
-  return (
-    <Link href={link}>
-      <button className={styles.Button}>{children}</button>
-    </Link>
-  );
+const Button = ({ children, link, target }) => {
+  if (target) {
```
imjuangarcia, line 6: I wonder if we should explicitly check here if `target === '_blank`, since there are [other target options](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target) besides `_blank` and `_self`.

I've hardly ever used them though, and they all behave like `_self`, so maybe it's not worth it?

components/blocks/button.js
imjuangarcia, line 8: Unrelated to this PR, so we can tackle it separately, but `<a>` tags should be used for links, and `<button>` tags for functionality, so using both here wouldn't be appropriate from an accessibility standpoint. I wonder if we can simplify this code like so:

```
if (target) {
    return (
      <a href={link} target={target} className={styles.Button}>
        {children}
      </a>
    );
  } else {
    return (
      <Link href={link}>
        <a className={styles.Button}>{children}</a>
      </Link>
    );
  }
```

components/blocks/button.js
```diff
@@ -2,12 +2,20 @@ import Link from "next/link";
...
+    return (
+      <a href={link} target={target}>
+        <button className={styles.Button}>{children}</button>
+      </a>
+    );
+  } else {
```
imjuangarcia, line 12: Feel free to ignore if you feel it hurts with the legibility of the code, but the `if...else` statement could be simplified by using a ternary operator, like so:

```
target ? <a href={link} target={target} className={styles.Button}>
    {children}
</a> : <Link href={link}>
     <a className={styles.Button}>{children}</a>
</Link>
```

components/navigation/arrowLink.js
```diff
@@ -42,6 +43,7 @@ const ArrowLink = ({ children, link, type, content }) => {
               ${styles.Link}
               ${styles.NextLink}
             `}
+            target={target ? target : "_self"}
```
imjuangarcia, line 46: SVGs don't need the `target` attribute

imjuangarcia (changes requested):
Most of my comments are nitpicks and/or optional things to check, the only one I'd address before merging the PR is removing the `target` attribute on the SVG in `ArrowLink`, as I believe it's not required. Other than that, looks good to me!

components/blocks/button.js
snehankekre: Oh good idea, that removes any ambiguity about what to set target to. Fixed on  ce80f8b.

components/blocks/button.js
snehankekre: All versions of major browsers now automatically use the behavior of rel="noopener" for any target="_blank" link, but I added it on b276d7c for best practice / convention, although it is redundant.

snehankekre: Really good point! Learned something new, and fixed on 2ed3f41 🙂

components/blocks/button.js
snehankekre: Doesn't hurt legibility much, so fixed on 0d1f981.

components/navigation/arrowLink.js
snehankekre: Yup, I didn't use it for the SVGs. The "Next: Read More ->" (see screenshot) use `arrowLink.js`. Currently, clicking them opens the links in the same tab. I add `target` to `arrowLink.js` and use that in news entry to open the blog posts in a new tab.

image

snehankekre:
@imjuangarcia  I didn't the `target` attribute in `ArrowLink` for the SVGs. The "Next: Read More ->" (see screenshot) uses `ArrowLink`. Currently, clicking them opens the links in the same tab. I add `target` to `ArrowLink` and use that in news entry to open the blog posts in a new tab.

image

imjuangarcia (approved):
Looks great, 🚀

PR #472, opened 2022-09-06
WIP: Add support for optional image in `newsEntry`

snehankekre:
## 🧠 Description of Changes

Adds support for an image in the `newsEntry` component used on the homepage to feature the latest blog posts.

TODO: @imjuangarcia Could use your help making it so that the date, title, text, and arrow link appear to the right of the featured image, in the same row. Here's what I mean:

image

**Revised:**

image

**Current:**

image

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small


imjuangarcia:
@snehankekre Updated the look of these items on [4144239](https://github.com/streamlit/docs/pull/472/commits/41442398eeb0a10a94b2579e3641c98b4869406c). Here's how they look like now 👇🏻 

Screen Shot 2022-09-06 at 2 01 14 PM

Also made some changes to the `newsEntry.js` file to avoid repetition. This line: `{image && <Image src={image} clean={true} />}` will only render if we have an `image` parameter, so we don't need to do an `if...else` condition 🙂 

Let me know if you need anything else from me here!

imjuangarcia (approved):
LGTM, but `blog-3.gif` is 3.5mb in size, so I'd recommend against using it, and replacing it with a raster image (JPG or PNG file) instead.

We're working on removing GIFs from our blog soon for the same reason, so it'll be good to keep consistent I believe!

snehankekre:
Look fabulous @imjuangarcia! 🤩 Thank you for the improvements. I replaced the large GIF with an SVG on ec74bbf.

PR #474, opened 2022-09-07
Update aws-s3.md with AWS_DEFAULT_REGION

### LINKED ISSUE #473: AWS S3 connection example breaks without AWS_DEFAULT_REGION
**Link to doc page in question (if any):** https://docs.streamlit.io/knowledge-base/tutorials/databases/aws-s3

**Name of the Streamlit feature whose docs need improvement:** AWS S3 connection (as documented in the provided example)

**What you think the docs should say:**

Configuring an AWS client with environment variables requires `AWS_DEFAULT_REGION` to be included or most clients will not fully recognize the creds (regardless of whether it's actually used). As written [here](https://docs.aws.amazon.com/sdkref/latest/guide/environment-variables.html) and in most other AWS docs on the subject.

This is not obvious, nor is it included in the current documentation for pulling data from an S3 bucket in Streamlit. Currently, using the example config verbatim causes the user to get a fairly opaque 403 for pulling data from a non-public bucket.

Suggested fix would be to just include this in the example. I am submitting a PR with this change. I considered whether or not to explicitly call this requirement out in the doc - didn't include that in my PR but maybe worth consideration.

JoshuaC215:
Add required AWS_DEFAULT_REGION to example secret values

## 📚 Context

See #473 

## 🧠 Description of Changes

Adds `AWS_DEFAULT_REGION` value to the included `secrets.toml` example with S3 creds

## 💥 Impact

Tiny, one line doc change

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

Resolves #473 


JoshuaC215:
Deploy failure appears to be unrelated to my change

```
8:51:26 PM: Starting build script
8:51:26 PM: Installing dependencies
8:51:26 PM: /opt/build-bin/run-build-functions.sh: line 223: /opt/buildhome/python3.7/bin/activate: No such file or directory
8:51:26 PM: Error setting python version from runtime.txt
8:51:26 PM: Please see https://github.com/netlify/build-image/blob/focal/included_software.md for current versions
8:51:26 PM: Build was terminated: Build script returned non-zero exit code: 1
8:51:26 PM: Creating deploy upload records
8:51:26 PM: Failing build: Failed to build site
```

snehankekre (approved):
Hi @JoshuaC215 👋 , thank you for catching this subtle bug, and for submitting a fix! I don't recall running into the opaque 403 when we published this tutorial last year, but I'm able to repro it now.

I've made a few additions to your PR:
- On 8282c00 I updated the screenshot to highlight the AWS region
- On 28ce8d0 added a note in step 1 to note down the AWS region for later. I also replaced `us-east-1` with a generic `xxx` in the secrets, as that might not always be the default region. I do call out that we use `us-east-1` in this example, but the user may have a different one.
- Pulled the changes from main in 3775e7e to fix the netlify builds.

Do you think we're good to merge? 😃

JoshuaC215:
@snehankekre thank you! One note, S3 buckets are technically global so it isn't strictly required that the provided AWS_DEFAULT_REGION matches the one the bucket was set up with. But probably useful to set it that way in the docs in case someone expands to another service (RDS etc) where region matters.

LGTM for merge :)

PR #485, opened 2022-09-13
Fix styling of inline code in headings

snehankekre:
## 📚 Context

I noticed that inline code in headings aren't properly styled. When inline code is included in headings, the font color of the `<code>` elements is not red and there is no margin and padding between the `<code>` element and the rest of the heading text, making the code appear as though it has been concatenated to the preceding and posterior text.

## 🧠 Description of Changes

- ~Adds the styling used for `p > code` to inline code in headings, but omits the gray borders as they look out of place in headings.~
- @imjuangarcia: wrap the `children` element in `headerLink` in a `<span>` when there's a `<code>` tag present, which helps with the spacing of words without having to rely on `px-1 mx-1`

**Revised:**

image

**Current:**

image

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 


components/utilities/headerLink.module.css
```diff
@@ -10,6 +10,10 @@
   @apply font-bold;
 }
 
+.HeaderContainer code {
+  @apply text-red-70 rounded-md px-1 mx-1;
```
imjuangarcia, line 14: I find the red text a bit too prominent in headings, especially with long commands:

Screen Shot 2022-09-13 at 08 05 33

I'd keep it black

components/utilities/headerLink.module.css
```diff
@@ -10,6 +10,10 @@
   @apply font-bold;
 }
 
+.HeaderContainer code {
```
imjuangarcia, line 13: Can we add classes in the HTML for the `<code>` tag? Or additional `<span>` elements inside the `<h1>`? We should find a way to only add the `px-1` and `mx-1` classes to `<code>` tags that are in the middle of the heading, like so:

```
<!-- These should have px-1 and mx-1 -->
<h1>I'm a nice `<code>` heading</h1>
```

And not to `<code>` tags that show up right after the `<h1>` tag, like so:

```
<!-- These code tags should only have margin at the right, since margin left will throw the alignment off -->
<h1><code>st.</code>experimental_singleton</h1>
```

See here:

Screen Shot 2022-09-13 at 08 05 33

imjuangarcia (changes requested):
Thanks for working on fixing this! I see two things on my end:

- I prefer to avoid red for the color on the title, since headings take much more space than inline code tags, and if feels too prominent in my opinion: Worth checking with @tvst to see what are his thoughts!
- We should find a way to add the margins on inline tags differently, since `px-1 mx-1` won't work for all use cases and would throw alignment off

components/utilities/headerLink.module.css
snehankekre: On second pass, I agree. The red is too much in headings. I've reverted to black on b81567e

components/utilities/headerLink.module.css
snehankekre: Really good points! I hadn't considered the consequences of blindly applying `px-1` and `mx-1` to headings. 

> Can we add classes in the HTML for the \<code> tag?

Where should this be done? In this same `.css` file or elsewhere?

> And not to \<code> tags that show up right after the \<h1> tag, like so

Same noob question as above. 😅  Can conditional application of `px-1` and `mx-1` based on the position of \<code> tags be done in CSS?

imjuangarcia: I was referring to turning the `.md` heading tag with the inline tag:

```
## Set up your `virtual` environment
```

into something like this on the HTML:

```
<h2><span>Set up your</span><code>virtual</code><span>environment</span></h2>
```

To make it easier to style, but that didn't work, so I ended up wrapping the `children` element in `headerLink` in a `<span>` when there's a `<code>` tag present, which helps with the spacing of words without having to rely on `px-1 mx-1`. Here's the result:

Screen Shot 2022-09-14 at 3 03 39 PM

Notice how `Prerequisites` is properly aligned, and `virtual` is spaced out from the rest of the words. Hope this helps!

snehankekre: > into something like this on the HTML
Ah, I didn't occur to me to try that! 

Your latest changes by wrapping the `children` element in `headerLink` in a `<span>` when there's a `<code>` tag present work well! Thanks for the more robust fix 😃

PR #483, opened 2022-09-12
Sticky navigation fixes

imjuangarcia:
## 📚 Context

This PR fixes a couple of issues reported by @tvst, where content would get hidden when the header was expanded/collapsed. See the attached Notion documents below for more insights.

## 🧠 Description of Changes

* Move the `isSticky` logic from `header.js` to `globalContainer.js`, and pass the `isSticky` value as a prop to both `<Header>` and the main children container;
* Update styles on both `sideBar.js` and `globalTemplate.js`, adding `padding-top` to accommodate for the change in height on the header.

**Revised:**

https://user-images.githubusercontent.com/34423371/189689115-6990e349-8c16-4d0e-8dee-0ba7b54e2cc4.mov

**Current:**

https://user-images.githubusercontent.com/34423371/189689318-251729f3-f01f-45a9-b341-ba52ad4b3c98.mov

https://user-images.githubusercontent.com/34423371/189689305-a00fd74b-351b-4da8-b317-3dd5da0e306a.mov

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

- https://www.notion.so/streamlit/Sidebar-when-header-is-collapsed-should-have-padding-to-hide-the-fade-effect-when-scrolled-to-the-to-151e32d21c494b8ebfc1325dea3bd327
- https://www.notion.so/streamlit/Docs-header-collapse-functionality-doesn-t-animate-and-hides-content-you-don-t-expect-it-to-be754ddc45c04906910da88b60dd8e5d

**Contribution License Agreement**

By submitting this pull request you agree that all contributions to this project are made under the Apache 2.0 license.

PR #490, opened 2022-09-16
Pass undefined to non-boolean attribute className in `globalTemplate.js`

snehankekre:
## 📚 Context

As part of Sticky navigation fixes (#483), the `isSticky` value is passed as a prop  to both `<Header>`  and the main children container. When passing it to the main children container, it was conditionally omitted using `&&`, which threw the following warning to console:

```
Warning: Received `false` for a non-boolean attribute `className`.

If you want to write it to the DOM, pass a string instead: className="false" or className={value.toString()}.

If you used to conditionally omit it with className={condition && value}, pass className={condition ? value : undefined} instead.
    at div
    at main
    at Layout (webpack-internal:///./components/layouts/globalTemplate.js:16:19)
    at Home (webpack-internal:///./pages/index.js:62:17)
    at AppContextProvider (webpack-internal:///./context/AppContext.js:13:31)
    at StreamlitDocs (webpack-internal:///./pages/_app.js:70:26)
    at StyleRegistry (/Users/skekre/docs/node_modules/styled-jsx/dist/index/index.js:671:34)
    at AppContainer (/Users/skekre/docs/node_modules/next/dist/server/render.js:415:29)
    at AppContainerWithIsomorphicFiberStructure (/Users/skekre/docs/node_modules/next/dist/server/render.js:447:57)
    at div
    at Body (/Users/skekre/docs/node_modules/next/dist/server/render.js:715:21)
```

## 🧠 Description of Changes

- Used the ternary operator instead and pass `undefined` if the condition is falsy.

**Revised:**

```js
<div className={isSticky ? styles.stickyPageWrapper : undefined}>
    {children}
</div>
```

**Current:**

```js
<div className={isSticky && styles.stickyPageWrapper}>{children}</div>
```

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small

### DIFF
--- a/components/layouts/globalTemplate.js
+++ b/components/layouts/globalTemplate.js
@@ -22,7 +22,9 @@ const Layout = ({ children }) => {
   return (
     <main id="root" className="dark:bg-gray-100">
       <Header isSticky={isSticky} />
-      <div className={isSticky && styles.stickyPageWrapper}>{children}</div>
+      <div className={isSticky ? styles.stickyPageWrapper : undefined}>
+        {children}
+      </div>
     </main>
   );
 };


imjuangarcia (approved):
Makes total sense, thanks for catching that!

PR #508, opened 2022-10-25
Migrate post build scripts to GitHub actions

imjuangarcia:
## 📚 Context

We uncovered a bug last week with our post-build tasks, where sometimes, when a build fails (even deploy previews) this can cause the Algolia search index to lose its configuration, thus breaking search in prod. To make sure this can’t happen, we're only triggering the postbuild scripts (building the search index, but also our sitemap) _only_ when merging to `main`, so we can ensure everything works as expected and the site build pass before running these scripts.

## 🧠 Description of Changes

* Removed the pre-push hook to generate the sitemap;
* Updated the way scripts are being run in `package.json`;
* Added GitHub action to run those scripts there.

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small 

## 🌐 References

- https://www.notion.so/streamlit/Algolia-doc-indexing-failure-in-docs-shouldn-t-break-the-search-index-63842ec273384e8e8b210e1e90a78802


snehankekre (approved):

scripts/build-search-index.js
```diff
@@ -11,21 +11,25 @@ const SKIP_THESE = ["/menu", "/404", "/500"];
...
-  fs.readdirSync(articleDirectory).forEach(function (file) {
-    const subpath = path.join(articleDirectory, file);
-    if (fs.lstatSync(subpath).isDirectory()) {
-      getAllFilesInDirectory(subpath, files);
-    } else {
-      files.push(subpath);
-    }
-  });
-  return files;
+  try {
```
sfc-gh-tteixeira, line 14: I don't think this does anything. Remove?

imjuangarcia: Removed on [8c6d551](https://github.com/streamlit/docs/pull/508/commits/8c6d551bf0316e3edf1d70c19e8c37b9e524d5a2)!

Issue #561, opened 2023-01-11
Improve your website with documentaion, please!


Arseni1919:
### Checklist

- [X] I have searched the [existing issues](https://github.com/streamlit/streamlit/issues) for similar issues.
- [X] I added a very descriptive title to this issue.
- [X] I have provided sufficient information below to help reproduce this issue.

### Summary

If I enter any documentation page in your website - https://docs.streamlit.io/library/api-reference - something strange is happenning - the page reloads itself several times, and it is very hard to go back therefore. The website also crashes because of this. I attached video that shoows the issue. Please, fix!

One minor thing: The "Contact Us" link in the footer of the website also does not help. It brings me directly to my Mail app on my Mac which I didn't configured (and don't want to). Just show your email explicitly on the website - and it will be perfect.
Thank you!

### Is this a regression?

- [ ] Yes, this used to work in a previous version.

### Debug info

- Streamlit version:
- Python version:
- Operating System:
- Browser:
- Virtual environment:

### Additional Information

https://user-images.githubusercontent.com/55444116/211796207-b9063ea5-7ca5-4430-949b-3072359ee99b.mov

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

snehankekre:
Thank you for reporting this issue @Arseni1919! We're aware that the site or your browser tab become unresponsive when specifically when viewing multiple API reference pages in succession. We're investing the issue and will let you know once it's fixed.

Arseni1919:
One minor thing: The "Contact Us" link in the footer of the website also does not help. It brings me directly to my Mail app on my Mac which I didn't configured (and don't want to). Just show your email explicitly on the website - and it will be perfect.
Thank you!

PR #562, opened 2023-01-11
Attempt to fix performance issue when browsing API reference page

sfc-gh-jgarcia:
## 📚 Context

Some folks had been experiencing slow load times and unresponsiveness for our docs site after browsing multiple API reference pages. This seems to be caused by the fact API reference pages have apps embedded on them, and they're not properly destroyed when the user abandons the page. This PR aims to fix that!

## 🧠 Description of Changes

* Added a unique `key` to each `<iframe>` added to the `<Cloud>` components, so it can be regenerated by React whenever it changes;
* Added a function to destroy old `<iframe>` tags, clone them, change its `src` attribute, and insert them back when examples are added through the `<Autofunction>` component.

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small 

## 🌐 References

- https://www.aleksandrhovhannisyan.com/blog/react-iframes-back-navigation-bug/#solution-remount-the-iframe-with-a-key
- https://www.notion.so/streamlit/Investigate-performance-issues-unresponsiveness-on-docs-site-3cbb0ba9d489483e9f41bc06b987ffe0
- https://github.com/streamlit/docs/issues/561

### DIFF
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -14,8 +14,6 @@ import "prismjs/plugins/toolbar/prism-toolbar";
 import "prismjs/plugins/copy-to-clipboard/prism-copy-to-clipboard";
 import "prismjs/plugins/normalize-whitespace/prism-normalize-whitespace";
 
-import useSourceFile from "../../lib/useSourceFile";
-
 import styles from "./autofunction.module.css";
 
 const cleanHref = (name) => {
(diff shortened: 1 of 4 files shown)


snehankekre (approved):
Amazing :heart_eyes: This seems to have fixed the issue. The site is so much more responsive and feels smooth. Thanks as always, @sfc-gh-jgarcia!

PR #565, opened 2023-01-12
Remove note on secondary background color for ploty

jrieke:
## 📚 Context

The docs state that we use the secondary background color as the background for Plotly charts. But we don't do this anymore with the new charting them. 

## 🧠 Description of Changes

- Removing the sentence. 
- Might be good to also update the screenshot but I don't have the underlying app. 

**Revised:**

(Leaving this out since it's a one-liner).

**Current:**

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

- [ ] [Notion](...)


snehankekre (approved):
You got to it before I could 😄 Thank you!

PR #568, opened 2023-01-17
Deprecation notice designs

sfc-gh-jgarcia:
## 📚 Context

This PR adds support to deprecate params across multiple places in our documentation site.

## 🧠 Description of Changes

* Add a `isDeprecated` boolean param that can be added to menu items in `menu.md`, to add an icon next to them in the side navigation;
* Add a `deprecated` boolean param that can be added to `<RefCard>` components to add the tooltip in reference card tiles;
* Styled `<blockquote>` tags inside `<RefCard>` components, so that every time you add one of them like so:

```markdown

<RefCard deprecated={true}>

> This command is deprecated.

</RefCard>

```

It'll rended the text inside the deprecation notice tooltip.

* Added `deprecated` and `deprecatedText` params to `<Autofunction>` components, so that every time you do something like this:

```markdown

<Autofunction deprecated={true} deprecatedText="This is my deprecation notice" />

```

It'll render the corresponding `<Deprecation>` component below the function title.

**Screenshots:**

_In side navigation_

Screenshot 2023-01-19 at 11 16 08 AM

_In API reference page_

Screenshot 2023-01-19 at 11 33 36 AM

Screenshot 2023-01-19 at 11 20 41 AM

_In function page_

Screenshot 2023-01-19 at 11 30 11 AM

_In function param_

Screenshot 2023-01-19 at 11 35 09 AM

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small 

## 🌐 References

- https://www.notion.so/streamlit/Designs-to-mark-API-reference-pages-or-parameters-as-deprecated-98e83a16439b4ad98022a5f0ed8a78a3

### DIFF
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -4,6 +4,7 @@ import classNames from "classnames";
 import Table from "./table";
 import { H2 } from "./headers";
 import Warning from "./warning";
+import Deprecation from "./deprecation";
 import { withRouter, useRouter } from "next/router";
 import Prism from "prismjs";
 import "prismjs/components/prism-python";
@@ -27,6 +28,8 @@ const Autofunction = ({
   streamlit,
   slug,
   hideHeader,
+  deprecated,
+  deprecatedText,
 }) => {
   const blockRef = useRef();
   const router = useRouter();
@@ -181,6 +184,13 @@ const Autofunction = ({
             </label>
           </form>
         </div>
+        {deprecated === true ? (
+          <Deprecation>
+            <p dangerouslySetInnerHTML={{ __html: deprecatedText }} />
+          </Deprecation>
+        ) : (
+          ""
+        )}
         <div
           className="code-desc"
           dangerouslySetInnerHTML={functionDescription}
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -224,7 +224,7 @@ const Autofunction = ({
       ? `
       <div class="${styles.DeprecatedContent}">
         <i class="material-icons-sharp">
-          info
+          delete
         </i>
         ${param.deprecated.deprecatedText}
       </div>`
@@ -234,7 +234,8 @@ const Autofunction = ({
       : `<p>No description</p> `;
 
     if (param.is_optional) {
-      row["title"] = `<p class="${isDeprecated ? "deprecated" : ""}">
+      row["title"] = `
+          <p class="${isDeprecated ? "deprecated" : ""}">
             ${param.name}
             <span class='italic code'>(${param.type_name})</span>
           </p> `;
@@ -243,7 +244,8 @@ const Autofunction = ({
         ${description}
       `;
     } else {
-      row["title"] = `<p class="${isDeprecated ? "deprecated" : ""}">
+      row["title"] = `
+          <p class="${isDeprecated ? "deprecated" : ""}">
             <span class='bold'>${param.name}</span>
             <span class='italic code'>(${param.type_name})</span>
           </p>`;
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -220,6 +220,15 @@ const Autofunction = ({
     const param = functionObject.args[index];
     const isDeprecated =
       param.deprecated && param.deprecated.deprecated === true;
+    const deprecatedMarkup = isDeprecated
+      ? `
+      <div class="${styles.DeprecatedContent}">
+        <i class="material-icons-sharp">
+          info
+        </i>
+        ${param.deprecated.deprecatedText}
+      </div>`
+      : "";
     const description = param.description
       ? param.description
       : `<p>No description</p> `;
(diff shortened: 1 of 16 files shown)


sfc-gh-jgarcia:
@snehankekre Here's the first draft of the front-end logic and styles for the deprecation notice stuff. Feel free to play with it and let me know what you think!

Would you be able to help with adding a `deprecated` boolean, or some way to identify deprecated params in the content that gets generated in the [`streamlit.json` file](https://github.com/streamlit/docs/blob/main/python/streamlit.json)? That way, I can use that in the [autofunction component](https://github.com/streamlit/docs/blob/main/components/blocks/autofunction.js) to cross out deprecated params

snehankekre:
@sfc-gh-jgarcia Amazing work 😍 !! All three of deprecation notices in `menu.md`, `<RefCard>` and `<Autofunction>` work perfectly. I can't state enough how great this is!

> Would you be able to help with adding a deprecated boolean, or some way to identify deprecated params in the content that gets generated in the [streamlit.json file](https://github.com/streamlit/docs/blob/main/python/streamlit.json)?

We've never deprecated params before but may in the future. When we do decide to deprecate a param, we'll have to add a `.. deprecated::` Sphinx directive in the docstring of the command in the core repo.  Let's say we want to deprecate the [`mime` param in `st.download_button`](https://github.com/streamlit/streamlit/blob/8909df7e0f232d8dbd2c02e64de81a5c32e7a386/lib/streamlit/elements/button.py#L162). The docstring would have to contain the deprecated directive like so:

```python
        """Display a download button widget.

        This is useful when you would like to provide a way for your users
        to download a file directly from your app.

        Note that the data to be downloaded is stored in-memory while the
        user is connected, so it's a good idea to keep file sizes under a
        couple hundred megabytes to conserve memory.

        .. deprecated:: 0.17.0
            ``mime`` was deprecated in version 0.17.0 and will be renamed to
            ``mime_type`` in version 2.0.0.

        Parameters
        ----------
        label : str
            A short label explaining to the user what this button is for.
            The label can optionally contain Markdown and supports the following
            elements: Bold, Italics, Strikethroughs, and Emojis.
```

I've updated our parsing logic on 743bb77 to check if the docstring contains this directive. If it does, we add a new [`deprecated`](https://github.com/streamlit/docs/blob/d7dd9d49d9e4f62192661a58f5d4e70111d87dc5/python/streamlit.json#L899) boolean field to the dict containing the param name in `python/streamlit.json`, and also add a [`deprecatedText`](https://github.com/streamlit/docs/blob/d7dd9d49d9e4f62192661a58f5d4e70111d87dc5/python/streamlit.json#L900) field whose value contains the deprecation text parsed from the directive.

Currently, I've added it for [`streamlit.download_button`](https://github.com/streamlit/docs/blob/d7dd9d49d9e4f62192661a58f5d4e70111d87dc5/python/streamlit.json#L898-L901).  Does that work and make sense?

snehankekre:
@sfc-gh-jgarcia Looks great to me! 🎉 

Btw, what's the reason we're not including the description text as a tooltip but rather above the param description? Just curious as the Figma mock showed the former:

image

sfc-gh-jgarcia:
> Btw, what's the reason we're not including the description text as a tooltip but rather above the param description?

Yeah, upon a bit of reflection, I felt the initial tooltip solution wasn't prominent enough, and I remember @sfc-gh-tteixeira mentioning the fact a deprecation notice should be "scary" so it's easy to find, understand what's going on, and avoid using it.

Happy to revert and use the tooltip though, if that's preferred: I'll circle back with Thiago regarding these designs later today, and make any adjustments you both might deem necessary!

snehankekre:
> I felt the initial tooltip solution wasn't prominent enough,

I concur. Best to be explicit about deprecation notices and not rely on users noticing the tooltip icon. 

Let's keep it as is (above the param description) unless Thiago has other suggestions.

sfc-gh-jgarcia:
@snehankekre Just updated the design tweaks @sfc-gh-tteixeira mentioned above. Let me know what you think!

The only thing I couldn't do _exactly_ as Thiago mentioned was the following:

> Padding in API page looks off. Place the icon inline with the text, to the right.

Since `<RefCard>` component's content is rendered via Markdown, so there's no way for me to place the icon inline with the title, since I can't edit the rendered output. I did update the icon placement to be inline with the title, though, which is something Thiago also suggested as an option.

Let me know what you both think, and if you need anything else from me here!

snehankekre:
> I did update the icon placement to be inline with the title, though

Fine by me!

snehankekre (approved):

snehankekre:
@sfc-gh-jgarcia Just noticed that with the latest changes, the deprecation notice tooltip + deprecation text when hovering over the trash icon on API reference pages is not displayed. The behavior from the screenshot below is missing:

image

e.g. neither of the following show the deprecation notice on hover:

```
<RefCard href="/library/api-reference/performance/st.cache" deprecated={true}>

<RefCard href="/library/api-reference/performance/st.cache" deprecated={true} deprecatedText="This param is deprecated.">
```

In its place, a light orange diamond is displayed on Trash icon hover:
image

PR #582, opened 2023-01-30
Make expanded state of `Collapse` component configurable via prop

snehankekre:
## 📚 Context

The `Collapse` component is used to expand/collapse content. There isn't a way to set the default expanded state when using the component.

## 🧠 Description of Changes

The following changes were made to the `Collapse` component:

- An `expanded` boolean prop was introduced to allow users to specify whether the content should be expanded or collapsed by default.
- A default value of `false` was set for the expanded prop in case it is not provided.

**Revised:**

```js
const Collapse = ({ title, children, expanded = false }) => {
  const [show, setShow] = useState(expanded);
  // ...
};
```

**Current:**

```js
const Collapse = ({ title, children }) => {
  const [show, setShow] = useState(false);
  // ...
};
```

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small

### DIFF
--- a/components/blocks/collapse.js
+++ b/components/blocks/collapse.js
@@ -2,8 +2,8 @@ import React, { useState } from "react";
 import styles from "./collapse.module.css";
 import classNames from "classnames";
 
-const Collapse = ({ title, children }) => {
-  const [show, setShow] = useState(false);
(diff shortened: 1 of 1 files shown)


sfc-gh-jgarcia (approved):
LGTM 👍

PR #622, opened 2023-04-14
`st.experimental_connection` API docs

snehankekre:

PR #623, opened 2023-04-20
Manually add components to API Reference

sfc-gh-jgarcia:
## 📚 Context

This PR adds some Streamlit components to our API Reference page

Screenshot 2023-04-20 at 10 19 30 AM

## 🧠 Description of Changes

**Revised:**

_Insert screenshot of your updated UI/code here_

**Current:**

_Insert screenshot of existing UI/code here_

## 💥 Impact

Size:

- [ ] Small 
- [ ] Not small 

## 🌐 References

- [ ] [Notion](...)


snehankekre:
The PR preview for the current changes look awesome! 😍 

Should we wait to a min of 3 components for `Developer tools`, `Connections and database`, and `Status elements` before merging? I'm onboard with merging now too, whatever you think is best.

Btw, I noticed the CTA is shown by default (before having reached the end of the slider). Just clarifying if this is intended behavior?

sfc-gh-jgarcia:
@snehankekre Thanks for the review! Find my comments below:
1. Just added a minimum of three components in those categories in [4a9db57](https://github.com/streamlit/docs/pull/623/commits/4a9db5775bbb39a315eeecde95ec0baead220700). Let me know what you think! My only question is that we don't have primitives (yet!) for two of those categories, and we're only showing third-party components... Is that desirable?
2. It is! I think it would look weird and produce a layout shift if it's not showing by default on the first three, but then it pops up after the first slide.
3. Thanks for catching that! Fixed it in [6827c37](https://github.com/streamlit/docs/pull/623/commits/6827c37feff04de6cc51a3b65b71655eee52b63f) by hiding the arrows in those sections where we have fewer than 3 components. Adding the loop in that case wouldn't make sense since there's no extra content to see.

snehankekre:
1. We can leave the connections category as-is. They will be added with primitives on #622 😄 As for showing components in categories for which do not have primitives, I'm not sure what we should do. Here are few options:
   - Easiest would be to omit the category and components. Cons: great components with a lot of usage are left out from docs (hampering discoverability).
   - Create the category with only components but don't create a page for the category and don't link to it in the sidebar nav. Pros: discoverability. Cons: could _appear_ as a bug that a page doesn't exist nor appears in the sidebar, slight possibility that users think they are primitives.
2. Rationale SGTM!
3. Awesome, makes sense :)

Unrelated to the above: I noticed the load time for the [API reference](https://deploy-preview-623--streamlit-docs.netlify.app/library/api-reference) page is ~5s when you navigate to it from the sidebar. Clicking on other pages results in those pages being rendered almost instantly. But if you click `API reference`, the slug changes instantly but the content takes ~5s to render, during which time the old page's content is shown. Perhaps a change from this PR has a noticeable performance cost? 🤔

snehankekre:
- The separate API pages for the components look great! Thank you!
> I'm ok with keeping components with no primitives on the general API reference page and not create a separated one 👍

 - That works 😄 
- Appreciate the deep dive into the performance issue and possible fixes! Certainly don't want to increase the Eng effort nor make it harder to update the components data. That said, given our API reference receives the most traffic (apart from Get started), we should IMO self-impose the constraint that this page feels snappy. It's definitely worth going over this with Thiago who may have pointers on how to find a middle ground here, such that the page feels snappy and we don't make the authoring experience + eng effort more complicated.

sfc-gh-jgarcia:
@snehankekre after reviewing this with @sfc-gh-tteixeira, we narrowed down the perf issue to the syntax highlighter. Performance for the API reference page significantly improved in [1b5c9da](https://github.com/streamlit/docs/pull/623/commits/1b5c9da9707f518c3394fc9d8a3319b36c5e8575) after temporarily removing syntax highlighting.

I initially thought that importing all the different languages in our `code.js` component, regardless of which we really need for each page, was causing the issue:

```
import "prismjs/components/prism-bash";
import "prismjs/components/prism-docker";
import "prismjs/components/prism-jsx";
import "prismjs/components/prism-python";
import "prismjs/components/prism-sql";
import "prismjs/components/prism-toml";
import "prismjs/components/prism-yaml";
```

But even after fixing this in [5a194b7](https://github.com/streamlit/docs/pull/623/commits/5a194b76d851fd0e0a81b4636717103423d4f347), the issue persists. So, the options I see here are, ranked from simpler to more complex:
* Don't do syntax highlighting for this page;
* Don't show code snippets for this page;
* Research and implement an alternative to Prism (our existing syntax highlighter);

Let me know what y'all think would be the preferred approach here, and we'll go from there. Thanks!

snehankekre:
@sfc-gh-jgarcia awesome that you and Thiago were able to narrow it down to the Prism syntax highlighter! It was very evident from [1b5c9da](https://github.com/streamlit/docs/pull/623/commits/1b5c9da9707f518c3394fc9d8a3319b36c5e8575) that page load significantly improved.  Of the options you've listed, not doing syntax highlighting is the easiest to do. Showing code snippets adds a lot of value to this page, but syntax highlighting doesn't comparatively add much.

However, there is another option I discovered over the weekend when running the Performance profiler from Chrome's DevTools.  We can improve our use of the highlighter to improve speed and continue syntax highlighting.

If you look at bottom-right cell of the below table, you'll see that most time is spent using prism's `highlightAll` function.  There are two specific improvements we can make:

1. The `highlightAll` function applies syntax highlighting to all matching elements in the DOM, which is be slow when there are many such elements. Instead, we can use the `highlightElement` function to only apply syntax highlighting to the current code block. We can pass the ref of the code element to this function.
2. We can use a memoization strategy to avoid reimporting the same Prism language component multiple times. For example, we can maintain a cache of already imported languages and only import a new language if it's not in the cache.

I've implemented both of the above on [bd7c517](https://github.com/streamlit/docs/pull/623/commits/bd7c51780e6a3c5974719897a43e9bbd04a355eb) and the page loads super quickly now, as shown in the left column below. LMK your thoughts on the same 😄 

| Revised | Current |
|:-------:|:-------:|
| Revised Summary | Current Summary |
| Revised Bottom-up View | Current Bottom-up View |

snehankekre:
Since Thiago specifically had comments, let's perhaps wait until he reviews or says otherwise in your next 1:1?

sfc-gh-jgarcia:
@snehankekre after my meeting with @sfc-gh-tteixeira yesterday, I've pushed [97c399c](https://github.com/streamlit/docs/pull/623/commits/97c399c97ccfd6e651b3825a67d0a346fea4e9ce) and [6a7fa0e](https://github.com/streamlit/docs/pull/623/commits/6a7fa0e60c37ae61fc4ec3cde4feca4bd4cf9030) with some visual tweaks.

The main change is removing the `Find more` CTA, because we felt it was repetitive to have it always present, in all categories, before the slider. Also, the content on the tooltip is already pointing folks to more resources, so we didn't think it was worth keeping it. 

Maybe what we can do is add a last slide on each category, pointing folks to the components Hub? What do you think? Other than that, feel free to merge if all looks good to you 👍

snehankekre:
Agreed that it wasn't worth keeping. Thanks for the visual tweaks!

> Maybe what we can do is add a last slide on each category, pointing folks to the components Hub

That's a good idea and something we can do in a follow-up PR 😄

PR #628, opened 2023-04-29
Fix ttl=None -> ttl=0 to disable caching in tutorials

sfc-gh-jcarroll:
## 📚 Context

Realized I had a brain fart last week and put in the wrong argument to disable caching in the two tutorials.

I scanned all the other docs and references, afaict this is the only place where I/we got it wrong. Also confirmed with Vincent that this new version is correct.

## 🧠 Description of Changes

- Switch `ttl=None` -> `ttl=0` for how to disable caching in the query methods of SQL and SnowparkConnection, as specified in the MySQL and Snowflake tutorials.

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

### DIFF
--- a/content/kb/tutorials/databases/mysql.md
+++ b/content/kb/tutorials/databases/mysql.md
@@ -93,7 +93,7 @@ for row in df.itertuples():
     st.write(f"{row.name} has a :{row.pet}:")
 ```
 
-See `st.experimental_connection` above? This handles secrets retrieval, setup, query caching and retries. By default, `query()` results are cached without expiring. In this case, we set `ttl=600` to ensure the query result is cached for no longer than 10 minutes. You can also set `ttl=None` to disable caching. Learn more in [Caching](/library/advanced-features/caching).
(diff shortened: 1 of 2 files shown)


sfc-gh-jcarroll:
cc @vdonato @snehankekre

snehankekre:
@sfc-gh-jcarroll The docstrings for the `query` [methods](https://deploy-preview-628--streamlit-docs.netlify.app/library/api-reference/connections/st.connections.sqlconnection#sqlconnectionquery), suggest the current version of setting `ttl=None` is correct. Could you confirm which is correct, @vdonato?

image

If setting `ttl=0` makes cache entries not expire, we should update the docstrings in core.

snehankekre (approved):
@sfc-gh-jcarroll The docstrings for the `query` [methods](https://deploy-preview-628--streamlit-docs.netlify.app/library/api-reference/connections/st.connections.sqlconnection#sqlconnectionquery), suggest the current version of setting `ttl=None` is correct; so does the type annot. Could you confirm which is correct, @vdonato?

image

image

If setting `ttl=0` makes cache entries not expire, we should update the docstrings in core.

sfc-gh-jcarroll:
Both the docstrings and these updates are correct:

- `ttl = None` : Cache the result, and don't ever expire the cache and don't re-run the function as long as the server is running
- `ttl = 0` : Don't cache the result, re-run the function every time

snehankekre:
Ah, thanks for clarifying! I re-read the docstrings and realised I was mistaken. `ttl=None` caches the query results indefinitely. This is not synonymous with _disabling_ caching.

PR #657, opened 2023-05-22
Handle prism module error and log to console instead of crashing app

snehankekre:
## 🧠 Description of Changes

- When inline code blocks in docs pages include a language that Prism does not support, it breaks the app.  This PR instead logs the error to the console instead of crashing the app

**Revised:**

image

**Current:**

image

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

- [ ] [Notion](...)


snehankekre:
Good call! I'll look for them, fix the imports and then merge.

PR #660, opened 2023-05-22
Media links update

MathCatsAnd:
## 📚 Context
Added CSS so that social media links in the footer would flex correctly in mobile view.

image


MathCatsAnd:
A proper web-dev might have a nicer fix for the template instead of just adding on a couple manual CSS lines, so no worries if there is a better way to do this. :)

snehankekre:
Looping in @sfc-gh-jgarcia who may be able to help with aligning the icons in the footer on mobile 😃

MathCatsAnd:
Since there is an odd number, I'm not sure there is much of a choice unless you want to give up centered, or force it to three rows instead of two upon flexing...

components/navigation/footer.module.css
```diff
@@ -28,6 +28,8 @@
 
 .SocialNetworks {
   @apply flex items-center gap-4;
+  flex-flow: row wrap;
+  justify-content: center;
```
sfc-gh-jgarcia, line 32: @MathCatsAnd I'd use Tailwind utility classes here, to keep it consistent with the rest of our codebase. So, instead of:

```css
.SocialNetworks {
  @apply flex items-center gap-4;
  flex-flow: row wrap;
  justify-content: center;
}
```

We could do:

```css
.SocialNetworks {
  @apply flex flex-wrap justify-center items-center gap-4;
}
```

And it should give us the same result 🙂

sfc-gh-jgarcia (commented):
Thanks for fixing this, @MathCatsAnd! Really appreciate it 🙏 

Your proposed fix is great: I think we need to have a broader conversation regarding whether we want to show these many social networks, and if so, then maybe we need to rethink the design a bit, but for now, you suggestion makes sense and addresses the underlying problem.

I'd only recommend swapping the vanilla CSS rules with Tailwind classes, to keep it consistent with the rest of our codebase. After that, I think we should be good to go here!

Let me know if you have questions 👍

MathCatsAnd:
I changed it to Tailwind syntax for the the flex in the footer icons.

@sfc-gh-jgarcia If you can clarify the syntax for custom colors, I can fix my other non-Tailwind overrides as well:

An existing example:
```css
.TwitterIcon {
  @apply hover:fill-darkBlue-60;
}
```
One of my manually set ones using a brand's hex color:
```
.DiscordIcon:hover {
  fill: #5865f2;
}
```

I didn't spend much time to try and learn Tailwind so I just figured out what worked by looking at Instagram's gradient background in plain CSS. I could identify "Streamlit red" as `red-70` but the obvious guesses of how to insert a hex color for Reddit, Discord, LinkedIn brand colors didn't work out.

sfc-gh-jgarcia:
@MathCatsAnd thanks again for updating this! See my comment below:

> I wonder if it's possible in Tailwaind to set some colors by other keywords than "red-50" and "blue-30" to instead create "brand-Discord" and "brand-Reddit" colors in the template

It is possible! See [here](https://tailwindcss.com/docs/customizing-colors#using-custom-colors) and [here](https://tailwindcss.com/docs/customizing-colors#arbitrary-values) for our specific use case. However, I think doing something like that would be a bit of an overkill for our use case, and also, I spotted another issue with this approach, where we have two different hover effects happening:

https://github.com/streamlit/docs/assets/103376966/0a993371-0f1a-43dc-bd74-5a3e66edc2ac

I'd simplify this and do something like this instead 👇 

https://github.com/streamlit/docs/assets/103376966/864cebd7-391c-41c4-bb2b-28dd61f70f72

Which is more subtle, but also consistent and independent on whatever we add there. @snehankekre what do you think?

The code to achieve my proposed hover state would be something like this:

```diff
 .IconLink {
-  @apply relative rounded-full overflow-hidden flex justify-center items-center px-2 sm:p-2 hover:opacity-100 hover:scale-105 hover:bg-gray-10;
+  @apply relative overflow-hidden flex justify-center items-center px-2 sm:p-2;
 }

-/* Colors for hover states */
-.GitHubIcon,
-.InstagramIcon,
-.MediumIcon {
-  @apply hover:fill-gray-90;
-}
-
-.YouTubeIcon {
-  @apply hover:fill-red-80;
-}
-
-.TwitterIcon {
-  @apply hover:fill-darkBlue-60;
-}
-
-.DiscordIcon:hover {
-  fill: #5865f2;
-}
-
-.RedditIcon:hover {
-  fill: #ff4500;
-}
-
-.LinkedInIcon:hover {
-  fill: #0077b5;
-}
-
-.NewsletterIcon:hover path {
-  stroke: black;
+.Icon:hover {
+  @apply opacity-70;
 }
```

MathCatsAnd:
@sfc-gh-jgarcia I noticed that,  too (how the hover colors the background first then the icon when you hover farther in). I had just tried to match what was there, but happy to remove it or correct it so the icon color is affected by the link hover making it in unison. 🙂

Just let me know. I'm not opposed to overkill if it's neat and tidy, but I defer of course.

sfc-gh-jgarcia:
Appreciated! Let's see how @snehankekre feels about simplifying the hover states, since he's the main DRI for all things documentation related 😄

MathCatsAnd:
For now I converted the existing behavior into Tailwind syntax (merging the `svg:hover` behavior into the `a:hover` behavior so it doesn't appear to be a two-stepper. Happy to delete the colors if the simplification is desired. (I was just here with itchy fingers, so...)
8894842F-9AE2-40C1-9725-3C36929C803F

snehankekre (approved):
@sfc-gh-jgarcia @MathCatsAnd I have to admit I'm not super confident with CSS and Tailwind classes yet 😅 

Approving this PR pending confirmation from @sfc-gh-jgarcia that the proposed changes are good to merge.

sfc-gh-jgarcia (approved):
LGTM, thanks for working on this, @MathCatsAnd !

PR #683, opened 2023-05-31
Fix `<iframe>` refreshing issue

sfc-gh-jgarcia:
## 📚 Context

This PR fixes an issue reported by @jrieke where embedded cloud apps were refreshing when the window was resized.

@snehankekre I did a bunch of manual testing already, but would be nice to do some more, checking responsiveness and whatnot, to ensure removing this "unused" code doesn't lead to any side effects

## 🧠 Description of Changes

* Removed some tech debt we had, an old function that calculated the window's `width` and `height`, presumably to be used in the sidenav.

**Revised:**

https://github.com/streamlit/docs/assets/103376966/c76941d2-8804-462c-9c1a-59d7a2bf6f20

**Current:**

https://github.com/streamlit/docs/assets/103376966/dd7fa453-b269-4056-afe2-7321d3343b2f

## 💥 Impact

Size:

- [] Small 
- [x] Not small 

## 🌐 References

- https://www.notion.so/snowflake-corp/Papercut-Request-iframes-refreshing-on-docs-page-d7429b40a19c4849985300a590828d1d?pvs=4

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -17,13 +17,13 @@ Router.events.on("routeChangeStart", () => NProgress.start());
 Router.events.on("routeChangeComplete", () => NProgress.done());
 Router.events.on("routeChangeError", () => NProgress.done());
 
-useEffect(() => {
-  if (navigator.platform.match("Mac") === null) {
-    document.body.classList.add("mac");
-  }
-}, []);
-
 function StreamlitDocs({ Component, pageProps }) {
+  useEffect(() => {
+    if (navigator.platform.match("Mac") === null) {
+      document.body.classList.add("mac");
+    }
+  }, []);
+
   return (
     <AppContextProvider>
       <Component {...pageProps} />


snehankekre (approved):
This issue has been bugging users for a long time. Kudos for the fix, @sfc-gh-jgarcia. I've verified the changes prevent the iframe refreshing issue.

From manual testing, I haven't spotted any unintended consequences yet when adjusting the size of the window and observing the behavior of different components. Since we don't have unit and integration tests setup yet, it's non-trivial to reliably find such issues.

That said, one way to test for unintended consequences is to check If the child component or any other component down the tree was relying on the window size information for their rendering or functionality. I think for now it's sufficient to confirm if the `window` prop passed down from `StreamlitDocs` is not used anywhere else. 

I wasn't able to find any other file importing `StreamlitDocs` and using the `window` prop. If this is the case, we should be more confident in merging these changes, right?

sfc-gh-jgarcia:
@snehankekre yeah, exactly! We checked the same thing with @sfc-gh-tteixeira and we couldn't find any place in the codebase where those properties were used.

snehankekre:
Great, I will merge now :shipit:

PR #708, opened 2023-06-18
change set_page_config arguments

astralko:
set_page_config does not allow "title" or "favicon"  but page_title and page_icon


snehankekre (approved):
Good catch, @astralko! Thank you for your contribution! Keep them coming 😄

PR #749, opened 2023-07-18
Feature: Add Escape key functionality to close enlarged images

snehankekre:
## 📚 Context

It's slightly annoying UX that one cannot close the enlarged image modal by pressing the Escape key, but have to click the close button on the modal's top right corner or click on the image itself (although the mouseover icon does not appear when enlarged, so it's not obvious users need to click the image to close it).

## 🧠 Description of Changes

- In the `<Image>` component, implemented a keydown event listener to detect when the `Escape` key is pressed
- Utilized the `useState` and `useEffect` hooks to open and close the modal and manage the addition and removal of the event listener

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

- [x] [Notion](https://www.notion.so/snowflake-corp/Docs-feature-Add-Escape-key-functionality-to-close-Image-component-258c5ec6cd5c470886509121acd10706)

**Contribution License Agreement**

By submitting this pull request you agree that all contributions to this project are made under the Apache 2.0 license.

PR #750, opened 2023-07-18
Shouldn't `onClick` event for close button be set to call `closeModal`, not `openModal`?

snehankekre:
## 📚 Context

When I was working on #749, I noticed in the `Image` component that the `onClick` handler for `CloseButton` calls the `openModal` function instead of `closeModal`. Shouldn't it be the opposite?

## 🧠 Description of Changes

- Sets the `onClick` event on the close button to call `closeModal` function which sets `isOpen` to `false`, hiding the modal.

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

- [ ] [Notion](...)

### DIFF
--- a/components/blocks/image.js
+++ b/components/blocks/image.js
@@ -37,7 +37,7 @@ const Image = ({ caption, pure, src, alt, clean }) => {
           {customCaption}
         </section>
         <section className={styles.LightBox} onClick={closeModal}>
-          <button className={styles.CloseButton} onClick={openModal}>
+          <button className={styles.CloseButton} onClick={closeModal}>
             close
           </button>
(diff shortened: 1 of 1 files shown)


sfc-gh-jgarcia (approved):
lol thanks for catching that! I guess, since the wrapping `<section>` element also has an `onClick` handler to close the model, it worked due to event propagation/bubbling, even though the handler was set incorrectly. All good to merge imo!

snehankekre:
@sfc-gh-dmatthews This branch doesn't contain the latest changes from `main`. Once merged, the functionality from this PR _and_ the Esc feature will work :)

PR #755, opened 2023-07-19
Version selector persists when st.command doesn't exist in version

snehankekre:
## 📚 Context

This PR makes the version selector in `st.command` API reference pages persist when the command didn't exist in that version. For example, if I go to an older version where the element didn't exist, the version selector disappears and I can't go back to the other versions.For example, for `st.experimental_connection`, if I navigate to [v1.21.0 or older](https://docs.streamlit.io/1.21.0/library/api-reference/connections/st.experimental_connection#stexperimental_connection), I'll see the warning that it didn't exist. But at this point, I could only use the back button of my browser to go back instead of continuing to use the version selector, which breaks the UX a little bit. 

The problem here is that when a command does not exist in a certain version, the component directly returns a `<Warning />` component without the version selector. To fix this issue, we now extract the version selector into a separate functional component and add it to both branches of our render.

## 🧠 Description of Changes

- Made the version selector independent of the existence of the `streamlitFunction` in `streamlit`. Even if the streamlitFunction does not exist, we show the version selector dropdown, allowing the user to select a different version where the `st.command` may exist. This is done by moving the version selector out of the header JSX, so it is rendered irrespective of whether `streamlitFunction` exists in `streamlit` or not.
- Replaced 'streamlit' Prefix in StreamlitFunction: When rendering the command name in the header when the command didn't exist in that version , 'streamlit' is replaced by 'st'. Currently for e.g., it says `streamlit.experimental_connection`.
- Added 'max-width: fit-content' to Form / dropdown style. @sfc-gh-jgarcia I could use your input here 😄 Without this change, here's what the form looked like:

image

I wasn't sure if there was a native tailwind equivalent to doing ` 'max-width: fit-content'`.

**Revised:**

image

**Current:**

image

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small 

## 🌐 References

- [x] [Notion](https://www.notion.so/snowflake-corp/Bug-version-selector-should-not-disappear-when-st-command-doesn-t-exit-in-that-version-3c16e8d0bd0a47929e5c06c38a4e3f51)

### DIFF
--- a/components/blocks/autofunction.module.css
+++ b/components/blocks/autofunction.module.css
@@ -15,8 +15,7 @@
 }
 
 .Form {
-  @apply relative my-4 md:my-0 ml-0 md:ml-auto;
-  max-width: fit-content;
+  @apply relative my-4 md:my-0 ml-0 md:ml-auto max-w-fit;
 }
 
 .Form::after {
--- a/components/blocks/autofunction.module.css
+++ b/components/blocks/autofunction.module.css
@@ -16,6 +16,7 @@
 
 .Form {
   @apply relative my-4 md:my-0 ml-0 md:ml-auto;
+  max-width: fit-content;
 }
 
 .Form::after {
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -96,9 +96,38 @@ const Autofunction = ({
     setIsHighlighted(true);
   };
 
+  const VersionSelector = ({
+    versionList,
+    currentVersion,
+    handleSelectVersion,
+  }) => {
+    const selectClass =
+      currentVersion !== versionList[0]
+        ? "version-select old-version"
+        : "version-select";
+
+    return (
+      <form className={classNames(selectClass, styles.Form)}>
+        <label>
+          <span className="sr-only">Streamlit Version</span>
+          <select
+            value={currentVersion}
+            onChange={handleSelectVersion}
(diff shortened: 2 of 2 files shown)


snehankekre:
Totally missed that, good catch Debbie

snehankekre:
Fixed the version selector position issue on 71e9abd

sfc-gh-dmatthews (approved):

sfc-gh-dmatthews:
Preview looks good to me! I will check that one off my docs list now. ✅ Thanks @snehankekre

components/blocks/autofunction.module.css
```diff
@@ -16,6 +16,7 @@
 
 .Form {
   @apply relative my-4 md:my-0 ml-0 md:ml-auto;
+  max-width: fit-content;
```
sfc-gh-jgarcia, line 19: @snehankekre We can use `max-w-fit` here as the Tailwind equivalent, see [here](https://tailwindcss.com/docs/max-width)

sfc-gh-jgarcia (approved):
LGTM, thanks for putting this together! Just added the Tailwind equivalent to `max-width: fit-content;`, but everything else looks good to go +1

components/blocks/autofunction.module.css
snehankekre: Fixed on 6a5bc20

PR #781, opened 2023-08-18
Fix scrollbars on MacOS

sfc-gh-jgarcia:
## 📚 Context

This PR fixes an issue reported by @sfc-gh-tteixeira where the Algolia popup would not have its right corners rounded.

Screenshot 2023-08-17 at 11 43 00

## 🧠 Description of Changes

The main reason for this was that some styles we added for our scrollbars, which were meant to be applied on Windows machines, were also "leaking" on Mac computers as well. The reason for this was that our code was checking for platforms that were called `Mac`, but now that Apple also has `MacIntel` platforms, we need to ensure we accommodate for both scenarios. Changes made where:
* Updated `pages/_app.js` to check if the word `Mac` is present on the `navigator.platform` prop, instead of strictly checking for it to be `Mac` only;
* Added some specificity to the `scrollbars.scss` `:not(.mac)` selector

**Revised:**

https://github.com/streamlit/docs/assets/103376966/8b82a3ae-8a5f-4a3a-b6c6-3f76aee9ddbf

**Current:**

https://github.com/streamlit/docs/assets/103376966/0ddcde60-c1fc-4aaf-a716-960e0a369b2a

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

- https://www.notion.so/snowflake-corp/Corners-in-Algolia-popup-should-be-round-84b2096022004766a3b55ff61a465702?pvs=4

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -19,7 +19,6 @@ Router.events.on("routeChangeError", () => NProgress.done());
 
 function StreamlitDocs({ Component, pageProps }) {
   useEffect(() => {
-    console.log(navigator);
     if (navigator.platform.includes("Mac")) {
       document.body.classList.add("mac");
     }
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -19,7 +19,8 @@ Router.events.on("routeChangeError", () => NProgress.done());
 
 function StreamlitDocs({ Component, pageProps }) {
   useEffect(() => {
-    if (navigator.platform.match("Mac") === null) {
+    console.log(navigator);
+    if (navigator.platform.includes("Mac")) {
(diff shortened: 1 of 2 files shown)


sfc-gh-dmatthews:
I'm not sure if it's intentional or not, but when I look at the preview, the scroll bar stays in light mode when the site is in dark mode (both for the search dialog and the main page). I looked in Chrome, Firefox, and Safari on MacOS.

image

sfc-gh-dmatthews (approved):

snehankekre (approved):
Thank you @sfc-gh-jgarcia!

PR #797, opened 2023-08-29
Add LLM glossary

sfc-gh-dmatthews:
## 📚 Context
Adding a glossary of terms. This is an initial placeholder which will be iterated on.

## 🧠 Description of Changes
Added a glossary targeted to LLMs within the Knowledge Base.

## 💥 Impact
- [X] Not small

## 🌐 References
Notion [blog to docs](https://www.notion.so/snowflake-corp/Glossary-of-LLM-terms-0f4f2bdc570a4cfb9c587bc9aaf0ed83?d=82d9bebbac73496fba94e61c1a6cadbf)

### DIFF
--- a/content/kb/tutorials/llm-apps/glossary.md
+++ b/content/kb/tutorials/llm-apps/glossary.md
@@ -29,21 +29,9 @@ GPTs are the superheroes of text generation. They’re LLMs that use the transfo
 
 Deep learning (DL) is a branch of ML where neural networks with multiple layers learn to perform complex tasks by progressively extracting higher-level features from data. It’s very good at image recognition, speech processing, and natural language understanding due to its ability to capture intricate data patterns.
 
-### **Transfer learning**
-
-Transfer learning is a technique where a model pre-trained on one task is fine-tuned for a different but related task. In LLM apps, transfer learning involves using a pre-trained LLM as a starting point and adapting it to specific tasks.
-
-### **Zero-shot learning**
-
-Zero-shot learning is a technique where a model is trained to perform tasks it hasn't seen during training. LLMs can generalize from prompts to generate responses for tasks they were not explicitly trained on.
-
 ### Few-shot learning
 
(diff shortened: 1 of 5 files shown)


sfc-gh-dmatthews:
@sfc-gh-jgarcia I added a second redirect to `netlify.toml` to override menu navigation like we do for /library -> /library/get-started. [Netlify's documentation](https://docs.netlify.com/routing/redirects/) suggests I can do this and I wanted to keep this kind of redirect separate from the usual redirects in `_redirects`. However, in the preview, the second redirect is not working. (The first one works, but the second one I added doesn't. I understand that neither work when I'm running locally.)

Any ideas?

snehankekre (approved):
From a v1 mvp perspective, the content LGTM. It could benefit from a callout somewhere at the top that this document is a work in progress, something that will grow over time as we add more terms.

sfc-gh-jgarcia:
> @sfc-gh-jgarcia I added a second redirect to `netlify.toml` to override menu navigation like we do for /library -> /library/get-started. [Netlify's documentation](https://docs.netlify.com/routing/redirects/) suggests I can do this and I wanted to keep this kind of redirect separate from the usual redirects in `_redirects`. However, in the preview, the second redirect is not working. (The first one works, but the second one I added doesn't. I understand that neither work when I'm running locally.)
> 
> Oddly, the second redirect works when I manually enter the URL, just not if I click on the item in the menu. This link to [LLM apps](https://deploy-preview-797--streamlit-docs.netlify.app/knowledge-base/tutorials/llm-apps) will redirect, but it doesn't function completely like the first redirect.
> 
> Any ideas?

@sfc-gh-dmatthews this doesn't work because we use [NextJS' `<Link />` tag](https://nextjs.org/docs/pages/api-reference/components/link) for these links, instead of normal `<a>` tags. `<Link />` tags offer many performance benefits, but they don't do a full page refresh like normal links do, and that's the reason why the redirect works when you manually enter the URL, but not when you click on the menu item. You can check the workaround we put together for the `/library` menu item [here](https://github.com/streamlit/docs/blob/main/components/navigation/navItem.js#L83).

PR #821, opened 2023-09-20
Update gsheet connection tutorials

sfc-gh-dmatthews:
## 📚 Context
Update connection tutorials to use st.experimental_connection

## 🧠 Description of Changes
* Public GSheet connection tutorial
* Private GSheet connection tutorial

## 💥 Impact
Size: Not small

## 🌐 References
Notion task: [Public GSheet Connection](https://www.notion.so/snowflake-corp/Connection-tutorial-Public-Google-Sheet-07cc6a7b16cb4006acaeb5e906ca82ce?pvs=4)
Notion task: [Private GSheet Connection](https://www.notion.so/snowflake-corp/Connection-tutorial-Private-Google-Sheet-6d530027277840d39de8fba870bd15d6?pvs=4)

### DIFF
--- a/content/kb/tutorials/databases/private-gsheet.md
+++ b/content/kb/tutorials/databases/private-gsheet.md
@@ -7,18 +7,29 @@ slug: /knowledge-base/tutorials/databases/private-gsheet
 
 ## Introduction
 
-This guide explains how to securely access a private Google Sheet from Streamlit Community Cloud. It uses the [gsheetsdb](https://github.com/betodealmeida/gsheets-db-api) library and Streamlit's [Secrets management](/streamlit-community-cloud/deploy-your-app/secrets-management).
+This guide explains how to securely access a private Google Sheet from Streamlit Community Cloud. It uses [st.experimental_connection](/library/api-reference/connections/st.experimental_connection), [Streamlit GSheetsConnection](https://github.com/streamlit/gsheets-connection), and Streamlit's [Secrets management](/library/advanced-features/secrets-management).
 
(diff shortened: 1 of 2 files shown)


content/kb/tutorials/databases/private-gsheet.md
```diff
@@ -116,36 +113,32 @@ Copy the code below to your Streamlit app and run it.
...
+conn = st.experimental_connection("gsheets", type=GSheetsConnection)
 
-sheet_url = st.secrets["private_gsheets_url"]
-rows = run_query(f'SELECT * FROM "{sheet_url}"')
+df = conn.read(
+    worksheet="Sheet1",  # The first worksheet is used if not specified.
+    ttl="10m",
+    usecols=[0, 1],
+)
+df.dropna(inplace=True)
```
snehankekre, line 126:
```suggestion
df = conn.read(
    worksheet="Sheet1",  # The first worksheet is used if not specified.
    ttl="10m",
    usecols=[0, 1],
    nrows=3,
)
```

How about using `nrows` rather than inplace operations we don't encourage due to Streamlit's execution model?

snehankekre (approved):
Approved. Can we confirm with Joshua if we should revise the _public_  gsheet example to use `st.experimental_connection`. It's much simpler without it, so I'm leaning towards not updating it at all.

content/kb/tutorials/databases/private-gsheet.md
sfc-gh-dmatthews: The funny thing is that null rows are supposed to be ignored by default and the extra operation shouldn't be needed at all. But in executing the code, sometimes it grabs the empty rows and sometimes it doesn't. There may be some specific version involved in switching the behavior, and may even be a bug, but I haven't pinpointed it yet.

PR #844, opened 2023-10-18
Show topmost Autofunction headers in floatingNav

sfc-gh-dmatthews:
## 📚 Context
Headers generated within Autofunction do not populate the right floatingNav element. This makes it less accessible when more than one Autofunction exists on a single page (such as when listing methods that can be used with mutable elements).

When floatingNav searches for headers to populate itself, it [expects a link to the header to be the first child](https://github.com/streamlit/docs/blob/e260fceebaac11cd550d38a54328b21e1fed430c/components/utilities/floatingNav.js#L41) of the header. These anchors are generated for header in markdown but not for Autofunction.

## 🧠 Description of Changes
This PR manually adds links under the topmost header in Autofunction as a temporary solution. It follows [another place where this link is manually added](https://github.com/streamlit/docs/blob/e260fceebaac11cd550d38a54328b21e1fed430c/components/blocks/headers.js#L24). It also prevents the first header from being removed unless it is an H1 header.

## 💥 Impact
Size: Not small

## 🌐 References
Notion Inbox: [Anchor navigation for Autofunction](https://www.notion.so/snowflake-corp/Anchor-navigation-for-Autofunction-8f47f04edc7340acb247bf1a4746f40a?pvs=4)


sfc-gh-jgarcia (approved):
Just added a minor style tweak, but other than that LGTM 👍

PR #845, opened 2023-10-18
Feature/menu updates

sfc-gh-dmatthews:
## 📚 Context
This PR adds two new feature to the main menu: divider lines and cross links. Divider lines allow for further sub-categorizing without creating additional nesting. Cross linking allows a page to appear in more than one place in the menu (for when users may be strongly inclined to look for it in more than one place e.g. `st.form` is container, though it's canonical location is with control flow). Cross linking is meant to be used lightly when an element is heavily torn between categories. This is to improve discoverability.

## 🧠 Description of Changes
Divider elements can be added by creating categories in menu.md without setting a URL property. These dividers can be labeled or unlabeled. Unlabeled dividers are achieved by setting the label to be "---"

Cross linking is accomplished by creating a link to a page using its full https:// address. Now, navChild.js will distinguish truly external links from local ones to adjust the icon accordingly. Additionally, once the cross link is detected, the URL is truncated to a relative one. This makes it functional in previews as well as preserves the efficiency of relative links. It also allows version compatibility.

**Revised:**
<table>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## 💥 Impact
Size: Not small

## 🌐 References
Notion inbox: [Menu dividers](https://www.notion.so/snowflake-corp/Docs-menu-divider-4683301485284f6c88875c56fd5366a8?pvs=4)
Notion inbox: [Cross linking](https://www.notion.so/snowflake-corp/Functionalize-redirect-workaround-used-for-library-a8e1e8f111b2412eaa520a30cb4446b9?pvs=4)


components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+  let url;
+  if (typeof page.url !== "undefined") {
+    url = page.url;
+  } else {
+    url = "";
+  }
 
-  const isLocalPage = page.url.startsWith("/");
+  const isLocalPage = url.startsWith("/");
+  const isCrossLinkedPage = url.startsWith("https://docs.streamlit.io");
```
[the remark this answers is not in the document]
sfc-gh-jgarcia: Logic here can be faulty, since we could also have URLs that start with `www` or `http` if you want to live on the edge... Maybe we should check if the url includes `docs.streamlit.io` instead. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
     target = "_blank";
   }
 
-  if (page.isVersioned && version && isLocalPage) {
+  if (isCrossLinkedPage) {
+    url = url.replace("https://docs.streamlit.io", "");
+    icon = <i className={styles.CrossLinkedIcon}>link</i>;
+  }
+
+  if (page.isVersioned && version && isLocalPage | isCrossLinkedPage) {
```
sfc-gh-jgarcia, line 92: I think there's an error here. Should be `if (page.isVersioned && version && isLocalPage || isCrossLinkedPage)`

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+  if (typeof page.url !== "undefined") {
+    url = page.url;
+  } else {
+    url = "";
+  }
 
-  const isLocalPage = page.url.startsWith("/");
+  const isLocalPage = url.startsWith("/");
+  const isCrossLinkedPage = url.startsWith("https://docs.streamlit.io");
+  const isDivider = url === "";
```
sfc-gh-jgarcia, line 80: It might be fine for now, but I feel it's a bit too hacky to fake a divider if the url is empty 😄 I feel it can lead to unexpected behavior in the future. Maybe we add a comment explaining why we're doing it this way, with a `TODO` to fix it later on?

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+
+  if (page.isVersioned && version && isLocalPage | isCrossLinkedPage) {
     // We need to version this URL, check if the URL has a version for this version
-    const newSlug = page.url.split("/");
+    const newSlug = url.split("/");
     newSlug[0] = version;
     url = `/${newSlug.join("/")}`;
   }
 
+  if (page.name == "---") {
```
sfc-gh-jgarcia, line 99: I'm probably missing something here, but if we have this check here, why do we need all the `isDivider` logic above?

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+    divider = (
+      <span className={styles.LinkContainer}>
```
sfc-gh-jgarcia, line 101: Can we use a `<div>` here instead of `<span>`?

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+        <div className={styles.DividerLine}></div>
```
sfc-gh-jgarcia, line 102: Can we use an `<hr />` here, instead of a `<div>`?

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+      </span>
+    );
+  } else {
+    divider = (
+      <span className={styles.LinkContainer}>
```
sfc-gh-jgarcia, line 107: Same comment here about changing the tag to a `<div>`

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+        <div className={styles.DividerLine}></div>
```
sfc-gh-jgarcia, line 108: Same comment here about using `<hr />`

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+        <div className={styles.DividerText}>{page.name}</div>
```
sfc-gh-jgarcia, line 109: Can we use a `<p>` or `<span>` tag here, instead of `<div>`?

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+        <div className={styles.DividerLine}></div>
```
sfc-gh-jgarcia, line 110: Can we use `<hr />` here?

components/navigation/navChild.module.css
```diff
@@ -17,6 +17,11 @@
 
 .ExternalIcon {
   @apply relative z-10 cursor-pointer text-sm ml-1;
+  @apply text-gray-70 !important;
```
sfc-gh-jgarcia, line 20: Should we be changing this color to something lighter for dark mode, or is contrast good enough?

components/navigation/navChild.module.css
```diff
@@ -40,6 +45,19 @@
   @apply text-base relative left-1 -top-[1px];
 }
 
+.DividerText {
+  @apply grow-0;
+  @apply text-gray-70;
```
sfc-gh-jgarcia, line 50: Same question about color for dark mode here

tailwind.config.js
```diff
@@ -41,6 +41,7 @@ module.exports = {
 
     fontSize: {
       // Defaults are: https://tailwindcss.com/docs/font-size
+      "2xs": ["0.625rem", { lineHeight: "1.25rem", letterSpacing: "0.15em" }],
```
sfc-gh-jgarcia, line 44: I don't think we should be adding a new class inside `tailwind.config.js` for this one-off scenario, which will be scarcely used. You could use Tailwind's arbitrary values instead: https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values

sfc-gh-jgarcia (changes requested):
Thanks for putting this together! Just left my comments on the PR, let me know if they make sense 👍

components/navigation/navChild.js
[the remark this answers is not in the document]
sfc-gh-dmatthews: Variable are renamed and will show in next commit.

I was thinking of requiring the full HTTP path since it makes the later task of converting it to a relative link simpler. The consequence of getting this wrong with a later content update is just that it would look and behave like an external link and the symbol would make the mistake visually apparent.

I can change it if you think it's better to allow both versions of the URL.

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
   }
 
   let link;
+  let divider;
   let icon;
   let target;
-  let url = page.url;
+  let url;
+  if (typeof page.url !== "undefined") {
```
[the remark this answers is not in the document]
sfc-gh-dmatthews: Update will show in next commit.

components/navigation/navChild.js
sfc-gh-dmatthews: I can change it if you'd prefer a new parameter. I thought this was slick since it avoids cluttering things up with more parameters and reduces the burden of remembering more parameters. Forgetting to set the URL would be an error since the menu expansions are dependent on the URLs being set. Such a mistake would be visually apparent, too.

If you prefer I create a new parameter, what sounds good? Just `isDivider = true` directly in menu.md?

components/navigation/navChild.js
```diff
@@ -65,24 +65,53 @@ const NavChild = ({ slug, page, color, className }) => {
...
+    url = "";
+  }
 
-  const isLocalPage = page.url.startsWith("/");
+  const isLocalPage = url.startsWith("/");
+  const isCrossLinkedPage = url.startsWith("https://docs.streamlit.io");
+  const isDivider = url === "";
 
-  if (!isLocalPage) {
+  if (!isLocalPage && !isCrossLinkedPage && !isDivider) {
```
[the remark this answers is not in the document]
sfc-gh-dmatthews: After renaming the variables I made a fourth `isExternal` if not the other three, then changed this to simply `if (isExternal)`. Update will show in next commit.

components/navigation/navChild.js
sfc-gh-dmatthews: Updated to double bar and put parentheses around (now named) `(isRelativePath || isAbsolutePath)`. Update will show in next commit.

components/navigation/navChild.js
sfc-gh-dmatthews: The divider may be labeled or unlabeled. I could logic gate this a little better, but the code basically computes the divider version and category version of each element, then chooses which to return at the end. When computing the divider version, it's labeled or unlabeled which is what this portion is doing.

The reason I did it this way was to preserve the variable names, I could either dispense with the variable divider and directly put the contents in link under three conditions:
1. isDIvider and page.name == "---"
2. else isDivider
3. else

I thought having the divider variable made it a little clearer what was happening. If I change from "compute divider and link then choose," would you prefer I rename link to navItem or something?

components/navigation/navChild.js
sfc-gh-dmatthews: Update will show in next commit.

components/navigation/navChild.js
sfc-gh-dmatthews: Update will show in next commit.

components/navigation/navChild.js
sfc-gh-dmatthews: Update will show in next commit.

components/navigation/navChild.js
sfc-gh-dmatthews: Update will show in next commit.

components/navigation/navChild.js
sfc-gh-dmatthews: Update will show in next commit.

components/navigation/navChild.module.css
sfc-gh-dmatthews: I picked a medium grey that I thought looked good in both dark and light mode, but happy to follow up with color tweaks. I even thought about using the category color for dividers at the second level and then grey at the third (innermost level) so there could be some distinction if a category had dividers as both sibling and child. (It looks okay in all grey, so happy for it to be a follow up unless you think the visibility is subpar as is.

Instead of forcing the color, I could increase transparency to force it to simply be a shade less pronounced in both light and dark mode. Just LMK your preference.

tailwind.config.js
sfc-gh-dmatthews: I'm struggling to get the arbitrary-property syntax correct. I've got it working with vanilla CSS instead, though.
```
  font-size: 0.625rem;
  line-height: 1.25rem;
  letter-spacing: 0.15em;
```

components/navigation/navChild.js
sfc-gh-dmatthews: Just realized: Did you want to change the original code as well? (I changed the new ones to div but left the original span it was copied from.)

sfc-gh-dmatthews:
@sfc-gh-jgarcia Here's the appearance with the updated commits, checked in light, dark, and mobile views:

<table>
<tr>
<td></td>
<td></td>
<td></td>
</tr>
</table>

components/navigation/navChild.js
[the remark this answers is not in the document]
sfc-gh-jgarcia: Got it! Yeah, I guess in order to fully automate this we should:
1. Create an array of potential absolute URLs;
2. Loop through it to see if any of those combinations (`http`, `https:` `www`, `https://www`, etc) fit the `page.url`;
3. If so, `.replace()` it;

Which might feel like a bit over-engineered for something like this. I'm fine with keeping it as it is, but maybe add a comment explaining why we require this specific format?

components/navigation/navChild.js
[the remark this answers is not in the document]
sfc-gh-jgarcia: Thanks! I think that's helpful 👍

components/navigation/navChild.js
sfc-gh-jgarcia: I see what you're saying, and ultimately is up to you, since you'll be updating the website constantly, so I'm fine with whatever makes your work easier 👍 

If we keep this logic though, I'd recommend adding a comment either on this component, and/or in `menu.md` whenever we use `url: ""`, to explain to other people exploring the codebase why the url is empty

components/navigation/navChild.module.css
sfc-gh-jgarcia: After playing with it for a bit, I think the text color works well! I did update the divider lines [here](https://github.com/streamlit/docs/pull/845/commits/2df3a8d0434e18548b33d5b77719877784d8d70c), to make them less harsh. Let me know what you think!

tailwind.config.js
sfc-gh-jgarcia: I see! We might be using an older version of Tailwind, where this function didn't exist. 

I replaced the `letter-spacing: 0.15em` with the almost identic `tracking-widest` alternative (`0.1em`), `line-height: 1.25rem;` with the `leading-5` Tailwind equivalent, and kept the font size value as vanilla CSS: You can check it [here](https://github.com/streamlit/docs/pull/845/commits/de04c4eb6b90a10404b44d980801fde92538496c). Let me know if that looks good to you!

sfc-gh-jgarcia (approved):
Made a few tweaks and left a few comments, mainly related to explaining a few of the logic specifics, but overall LGTM 👍 

Thanks for working on this!

components/navigation/navChild.module.css
sfc-gh-dmatthews: LGTM

tailwind.config.js
sfc-gh-dmatthews: LGTM

components/navigation/navChild.js
sfc-gh-dmatthews: I added a comment to the component.

components/navigation/navChild.js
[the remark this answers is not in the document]
sfc-gh-dmatthews: I added a comment to the component.

sfc-gh-dmatthews:
I've added the comments. I'm also happy to follow up on this with further improvements, but it will be convenient to have it available sooner for some other projects. Thanks so much!

PR #866, opened 2023-10-28
Use `overflow-wrap: anywhere` for headers

MathCatsAnd:
Allow headers to soft wrap if possible and only break words if an otherwise unbreakable string overflows on a line by itself.

## 📚 Context
Long function signatures were causing unintended overflow on mobile view. This was fixed by introducing `overflow-wrap: break-word`, but this will prefer breaking a word over an available soft wrap. `overflow-wrap: anywhere` will utilize soft wrap when possible. [It is well-supported.](https://caniuse.com/mdn-css_properties_overflow-wrap_anywhere)

**Revised:**
image

**Current:**
image

Words that are too long to fit on one line remain the same:
image

## 💥 Impact
Size: Small

### DIFF
--- a/components/utilities/headerLink.module.css
+++ b/components/utilities/headerLink.module.css
@@ -3,7 +3,8 @@
 }
 
 .HeaderContainer {
-  @apply flex items-center font-bold text-gray-90 leading-snug break-all md:break-normal;
+  @apply flex items-center font-bold text-gray-90 leading-snug;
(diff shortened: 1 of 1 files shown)


sfc-gh-jgarcia (approved):
Thanks for putting this together @MathCatsAnd! It's certainly a better solution than mine, all good to merge 👍

PR #888, opened 2023-11-29
Autofunction: Prevent method signatures with nested parentheses from truncating

sfc-gh-dmatthews:
## 📚 Context
When a class method signature contains nested parentheses, Authofunction incorrectly truncates the string "(...(...)"

## 🧠 Description of Changes
The regular expression used to get the method arguments now identifies the first `(` and last `)` instead of the first `(` and first `)`.

A temporary fix which replaced the nested right parent with `&rpar;` is reverted.

**Revised:**
image

**Current:**
(Without temporary fix)
image

## 💥 Impact
- [X] Small

## 🌐 References
- Notion task: [Fix nested parentheses in method signatures](https://www.notion.so/snowflake-corp/Autofunction-Fix-nested-parentheses-in-method-signatures-4a09ee81615b4c079ce86e378f21679c?pvs=4)

### DIFF
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -339,7 +339,7 @@ const Autofunction = ({
       .replace("streamlit", "st")
       .replace(/[.,\/#!$%\^&\*;:{}=\-`~()]/g, "");
     const type_name = method.signature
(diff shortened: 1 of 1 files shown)


snehankekre (approved):
LGTM 👍  Thanks for fixing the issue!

PR #901, opened 2023-12-18
Allow Streamlit in Snowflake versions in version selector

sfc-gh-dmatthews:
## 📚 Context
In order to document distinct behavior within Streamlit in Snowflake, this PR allows non-numeric versions beginning with "SiS" to be selected.

## 🧠 Description of Changes
Where the slug is inspected for a leading numeric component, an additional check is performed for a leading component beginning with "SiS" and possibly including numeric characters thereafter. Currently, there is just one SiS version but this future-proofs for the ability to document more than one SiS version.

The message to warn about looking at an older version and the message to alert if a function is not within the selected version were updated to reflect Streamlit in Snowflake instead of an "old" version.

**Revised:**
image

**Current:**
image

## 💥 Impact
Size:
- [X] Not small

## 🌐 References
[Notion project](https://www.notion.so/snowflake-corp/Docs-SiS-version-33a056f15619465c9ca148dd795cb1df?pvs=4)


sfc-gh-dmatthews:
This PR includes a temporary commit to modify streamlit.json that should be reverted before committing. I has a "SiS" version including only `st.write` for demonstration purposes.

sfc-gh-jgarcia (approved):
LGTM 👍

PR #904, opened 2023-12-22
Custom class guide

Asaurus1:
## 📚 Context

https://github.com/streamlit/streamlit/pull/7408 added Enum coercion to streamlit as a partial fix for a class of commonly-reported issues on the Streamlit bug trackker having to do with un-intuitive interactions between Streamlit's widget state, Python class equality and identity for user-defined classes, and the Streamlit execution model. 

As noted in discussions on [#7408](https://github.com/streamlit/streamlit/pull/7408), the Enum coercion fix is ultimately only able to automatically resolve only some types of these issues -- those involving Enums. Similar problems exist with other user defined classes such as `dataclasses` or resource managers.

This change adds a new User Classes page which provides advice to users on how to avoid the most common pitfalls of this class of problem, as well as explaining more in detail what the Enum coercion mechanic introduced by [#7408](https://github.com/streamlit/streamlit/pull/7408) actually does.

## 🧠 Description of Changes

* Added a user_classes.md file
* Updated configuration.md and the menu config to point to it.

**Revised:**

image

**Current:**

Page does not exist :)

## 💥 Impact

Visible to all users in documentation.

Size:

- [ ] Small 
- [x] Not small 

## 🌐 References

- https://github.com/streamlit/streamlit/pull/7408


sfc-gh-dmatthews:
Thanks @Asaurus1! Since we're just heading into the holiday break, it might take me a few weeks to look at this closely. A quick note off-hand: In the Streamlit docs, a "user" is someone viewing an app as opposed to the person developing an app, often "you." (At least, that's the intended standard. 😄) We'll probably want to avoid "user class" or "user-defined class" since it feels a little off for that reason. Maybe we'll go with "developer-defined classes" or "custom classes" but feel free to share your thoughts, too. I'm excited to dig into this. Happy holidays!

sfc-gh-dmatthews (changes requested):
This is fantastic! Thank you so much for the contribution! 🎉  I loved it.

Most of my comments fall into a few categories:
* Style, consistency, conventions
* Simplification or clarity
* Semantic headers

Feel free to let me know if you think any of my feedback is off or you want to go about it differently. I tried to explain why I was suggesting different things, but also feel free to ask any questions.

I put most things in as suggestions, but I can also directly edit if you didn't want to deal with revisions. Just let me know. Thanks again; this is going to be an awesome addition.

content/library/advanced-features/custom-classes.md
```diff
@@ -0,0 +1,150 @@
...
+    def __eq__(self, other):
+      # Two instances of MyDataclass are equal if both of their
+      # fields match.
+      return (self.var1, self.var2) == (other.var1, other.var2)
+
+  st.write(MyDataclass(1, 5.5) == MyDataclass(1, 5.5))
+  ```
+
+  The default python `__eq__` implementations for both regular
+  classes as well as `@dataclasses` depend on the in-memory ID of the class or class instance. This can cause comparison problems where two
```
blackary, line 84: This is true for regular classes, but not for dataclasses. It is certainly possible that you would want to ignore certain fields, and so the default `__eq__` might not always be appropriate, but much of the time it works great, because it does exactly this.

> eq: If true (the default), an [__eq__()](https://docs.python.org/3/reference/datamodel.html#object.__eq__) method will be generated. This method compares the class as if it were a tuple of its fields, in order. Both instances in the comparison must be of the identical type.
(From https://docs.python.org/3/library/dataclasses.html#module-contents)

```python
In [2]: @dataclass
   ...: class A:
   ...:     val: int
   ...:     other_val: int
   ...:

In [3]: a = A(1,2)

In [4]: b = A(1,2)

In [5]: c = A(1,3)

In [6]: a == b
Out[6]: True

In [7]: a == c
Out[7]: False
```

content/library/advanced-features/custom-classes.md
```diff
@@ -0,0 +1,150 @@
...
+  dataclass instances that you think should be equal are not evaluated
+  as being equal.
+
+- Alternatively, if you are storing data in `st.session_state`,
+  consider defining serialization and deserialization methods like `to_str` and `from_str` for your class, and use these to store class instance data in `st.session_state` rather than storing the class instance itself.
```
blackary, line 89: Can you explain why this is necessary or helpful? It might be, but I'm not sure when, as session_state works perfectly well to store whole dataclass objects. If you do need to serialize, then I would bias towards using the `asdict` or `astuple` methods from the dataclass module.

content/library/advanced-features/custom-classes.md
sfc-gh-dmatthews: @blackary Maybe I'm missing a detail, but I think to illustrate the situation here and why all this is a thing, try this:
```python
from dataclasses import dataclass

@dataclass
class A:
    val: int
    other_val: int
a = A(1,2)

@dataclass
class A:
    val: int
    other_val: int
b = A(1,2)

a == b #False
```
Between script runs, if the dataclass gets redefined then instances created on different reruns won't be equal even if instantiated with the same values.

Asaurus1:
@sfc-gh-dmatthews thanks! Happy to contribute. It's been a busy weekend but I should have some time to work on this in the next few days. You're also welcome to make any corrections you see fit -- I just don't know if that screws with the approval process having a committer also being a reviewer.

content/library/advanced-features/custom-classes.md
blackary: Ah! That makes sense, @sfc-gh-dmatthews. If you follow the recommendation of putting your class in a separate module, won't that eliminate this issue?

I can see that issue in the code below with class B, defined locally, but not with imported class A

```python
import streamlit as st
from dataclasses import dataclass
from modules.example import A

if "items" not in st.session_state:
    st.session_state["items"] = [A(1, 2)]


if st.button("Add another"):
    st.session_state["items"].append(A(1, 2))

st.write(st.session_state["items"])

for i in st.session_state["items"]:
    st.write(i)
    st.write(i == st.session_state["items"][0]) # ALL TRUE


@dataclass
class B:
    val: int
    other_val: int


if "items2" not in st.session_state:
    st.session_state["items2"] = [B(1, 2)]

if st.button("Add another 2"):
    st.session_state["items2"].append(B(1, 2))

st.write(st.session_state["items2"])

for i in st.session_state["items2"]:
    st.write(i)
    st.write(i == st.session_state["items2"][0]) # ALL FALSE AFTER FIRST ONE
```

sfc-gh-dmatthews: Yes, this whole page is about all the different ways to deal with the issue. The first, most-recommended option is to define classes outside the script and import them. It's not shown in the preview yet, since I put them in as suggestions in my other comments, but I'm recommending using headers "Option 1: ...", "Option 2: ....", "Option 3: ..." etc to emphasize that the guide is describing different alternatives to solving the one problem.

sfc-gh-dmatthews:
@Asaurus1 I've got lots of people around here who can bestow upon us the coveted green check, so no worries there. 😁 I'll leave the PR with the comments for the rest of the week and circle back this coming Monday. I can take care of anything left over or whatever you didn't have time to get to. Thanks!

content/library/advanced-features/custom-classes.md
Asaurus1: Yeah, I am working to see if I can explain the problem a little better. Basically if you do something like 
```
@dataclass
class MyDataclass:
   value:  str

if "foo" in st.session_state and st.session_state["foo"] == MyDataclass("bar"):
   st.success("you win")
if st.button("click me"):
   st.session_state["foo"] = MyDataclass("bar")
```
you'll never see that `st.success` because session_state has an old instance of MyDataclass in it. The suggestion to add a serialization method is only if you don't want to define a custom `__eq__` or put your class in a separate module file, hence why it's marked as "Alternatively";

As a note: even putting the class in a separate module file is not a fool-proof way out of this problem, it only reduces the frequency with which it can occur to "times when streamlit reloads modules, e.g. when someone edits a bit of page code". So the problem can still pop up while you're developing pages if the only mitigation you took is moving the class into a separate file. @sfc-gh-dmatthews do you think it makes sense to mention that in here or is that going too deep?

sfc-gh-dmatthews: I definitely think it's worth mentioning as "during development" is exactly when people are going to be encountering this and getting confused.

sfc-gh-dmatthews:
Sorry I couldn't get back to this last week because of the release, but I'm circling back now. 😄

sfc-gh-dmatthews:
Ok! I've just finished reviewing and editing everything. This really is great and I love your diagram; it's so pretty! However, I trimmed it out with a chunk of content in the "Class Redefinition" section since we want to avoid getting too far into the weeds. So even though you had a lot of really good things written out, I did condense that part.

There's a plan to have a developer-focused wiki in the future; that may be a good place for some deeper exposition. I just wanted to make sure there was a clear focus on the solutions here. I updated the general guidelines/patterns to try and have a consistent "broken" example that gets "fixed." The readers should be able to execute the snippets to see how each one does (or doesn't) work to help make it clear.

I'll just be giving it another proof read in the morning so this should be merged within a half a day.

Thanks again for the wonderful contribution!

Asaurus1:
Great! Thanks so much for the help.

sfc-gh-dmatthews (approved):

Asaurus1:
@sfc-gh-dmatthews love the edits <3 -- they make it so much clearer than my initial draft. Looks great!

Asaurus1:
@sfc-gh-dmatthews something is funky with one of the titles on the page on my mobile devide. 

Samsung Galaxy S21, Opera browser v 80.4.4424
snapshot_docs streamlit io_1709227071439

sfc-gh-dmatthews:
Thanks for reporting @Asaurus1. I've moved the comment to a new issue. It looks like the code text is displaying as a block so I'll take a look at that today.

PR #925, opened 2024-01-10
Refactor `RefCard` and `ComponentCard` to resolve React hydration errors

snehankekre:
## 📚 Context

This PR addresses React hydration errors caused by nested anchor tags in the `RefCard` and `ComponentCard` components.

## 🧠 Description of Changes

 - ComponentCard:
     - Implements conditional rendering to use a `div` with an `onClick` handler for navigation when children contain `<a>` tags. Retains the original `<a>` tag wrapper when no child anchor tags are present.
- RefCard:
    - Similar conditional rendering approach as `ComponentCard`, switching between a `div` and Next.js's `Link` component based on the presence of child anchor tags.
- Added a utility function (`scripts/contains-anchor.js`) to check if children contain an anchor tag and used in both updated components above.

**Revised:**

image
image

**Current:**

image
image

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small 

## 🌐 References


sfc-gh-dmatthews:
Thiago and I have talked about this PR (but not sure when it will get resolved exactly). I removed the final commit and merged in main to make it current and avoid conflicts. If Thiago goes with this approach, we can update the headers at the end or as a follow. :)

snehankekre:
I was just trying to edit some docs locally after months and I found the hydration errors painful when refreshing the API reference page. This PR makes the development experience better like #914. Any chance we could merge this or get an idea of what the blockers are? 😄

PR #937, opened 2024-01-20
Create oh-no-error-running-app.md

dataprofessor:
Created a self-serve post for a commonly encountered "Oh no" error.

## 📚 Context

The "Oh no" error is a commonly encountered error in the community. This consolidated self-serve post helps to guide users on resolving this issue on their own.

## 🧠 Description of Changes

This post borrows existing contents while also adding additional information to provide a thorough treatment of the topic.


sfc-gh-dmatthews:
Thanks for this! If I don't get to this by the end of the week, I'll do it next week.

I need to double-check the current state of the error screens as they were updated to show a different error page when resource limits are reached. I just want to make sure which conditions lead to which error screen since the plan is to have a complete list of all error messages with diagnostic steps and explanations.

image

PR #953, opened 2024-02-06
Add left and right margins for code in headers

sfc-gh-dmatthews:
## 📚 Context
When code text appears in headers, it does not have proper margins on the left and right. This PR adds the left and right margins.

## 🧠 Description of Changes
**Revised:**
image

**Current:**
image

## 💥 Impact
Size:
- [X] Small

### DIFF
--- a/components/utilities/headerLink.module.css
+++ b/components/utilities/headerLink.module.css
@@ -11,6 +11,10 @@
   @apply font-bold;
 }
 
+.HeaderContainer code {
+  @apply ml-1 mr-1;
+}
+
 h1.HeaderContainer {
   @apply leading-tight;
 }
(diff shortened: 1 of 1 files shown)


sfc-gh-dmatthews:
As noted in Slack, there is some [more complicated logic](https://github.com/streamlit/docs/blob/main/components/utilities/headerLink.js#L33C1-L37C5) for inline code within headers but I'm not sure if it ever triggers. (And if it does trigger, it looks like it would only do anything with the first instance of code within a given header.) I'm happy to review that more thoroughly at a later date and possibly remove that unused code if appropriate.

components/utilities/headerLink.module.css
```diff
@@ -11,6 +11,10 @@
   @apply font-bold;
 }
 
+.HeaderContainer code {
+  @apply ml-2 mr-2;
```
sfc-gh-arozanski, line 15: I guess you can also use mx-2

sfc-gh-arozanski (approved):

PR #958, opened 2024-02-08
Check for and show keyword-only arguments in Autofunction

sfc-gh-dmatthews:
## 📚 Context
Keyword-only arguments have had inconsistent treatment. A previous PR removed the inconsistent message of "This argument can only be supplied by keyword" with the intention of programmatically flagging them to be consistent and complete.

## 🧠 Description of Changes
* `generate.py` included extra lines to flag arguments a keyword only by adding a `is_kwarg_only` field.
* Autofunction will bump arguments from "Parameters" section to a "Keyword-only parameters" section with `is_kwarg_only` is true.
* Autofunction has also been corrected to meaningfully ignore methods defined outside of the Streamlit library. (Previously, a bug was exploited in Python 3.11 to discard these. It's now functional in Python 3.9.)
* If `is_kwarg_only` isn't present (as with previous versions until `streamlit.json` is updated), docstrings will display unchanged. Historical versions can be updated independently to flag keyword-only arguments.
* Removed unused deprecation logic. We do not have arguments within functions marked as deprecated since we save different docstrings for each version. Deprecation is noted at the function level. (We can re-add if we start to utilize that functionality.)

**Revised:**
<table>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
</table>

**Current:**
<table>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
</table>

## 💥 Impact
Size:
- [X] Not small

## 🌐 References
Notion task: [Improve keyword-only parameters in docstrings](https://www.notion.so/snowflake-corp/Improve-keyword-only-parameters-in-docstrings-30e899de60b34c09a31711a423e00031?pvs=4)

### DIFF
--- a/components/blocks/autofunction.js
+++ b/components/blocks/autofunction.js
@@ -160,6 +160,7 @@ const Autofunction = ({
 
   const footers = [];
   const args = [];
+  const kwargs = [];
   const returns = [];
   const versionList = reverse(versions.slice());
   let functionObject;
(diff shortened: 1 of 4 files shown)


snehankekre:
>We do not have arguments within functions marked as deprecated since we save different docstrings for each version.

We might want to keep it around. We're planning on deprecating an arg in `st.image` : https://www.notion.so/snowflake-corp/Spec-b8d929b452994869a8f6bbb693df2582

snehankekre:
It might be helpful to spot any bugs if we also regenerate `streamlit.json` for 1.31.0 and push it here.. To compare differences in docstrings here vs prod

snehankekre:
This is really awesome and something we've wanted to do forever. Fantastic stuff! We should run a quick design pass on whether we should include a separate section or the alternative styling you shared on Slack. I like the former, as it's super clear. But others may differ in preference.

sfc-gh-dmatthews:
@snehankekre Parameter-level deprecation formatting has been restored to autofunction.js. Also, this PR now includes the regenerated 1.31.0 docstrings. (They are parsed from 1.31.1 because of a missing docstring PR and include a manual edit to st.map and st.pydeck_chart for legal that will otherwise be reflected in 1.32.0.)

sfc-gh-tteixeira (approved):
Made a couple of [design tweaks](https://github.com/streamlit/docs/pull/958/commits/b8e834714f9d42867593efabf9c7f9162e41119a).

With those, this LGTM

PR #1025, opened 2024-04-04
IA rework to create Develop and Deploy categories

sfc-gh-dmatthews:
## 📚 Context
This PR changes "Streamlit library" into "Develop" and changes "Streamlit Community Cloud" into "Deploy."

## 🧠 Description of Changes
The new IA moves away from a product-focused organization to a task-focused organization. Community Cloud is presented as one option among many for how to deploy apps with additional pages added to generally describe deployment concepts.

Within Develop, the API reference was lightly rearranged for better grouping. "Advanced features" became "Concepts" and was heavily reorganized to group related concepts. Relevant content from the Knowledge base was elevated to Develop and Deploy. Tutorials are featured prominently now.

**Revised:**
image

**Current:**
image

## 💥 Impact
Size:
- [X] Huge

## 🌐 References
Notion project: [Docs: Develop, Deploy review](https://www.notion.so/snowflake-corp/Docs-Develop-Deploy-review-dac979882b214e23bd4b5102916b3477?pvs=4)

Closes #462 

**Contribution License Agreement**
By submitting this pull request you agree that all contributions to this project are made under the Apache 2.0 license.

PR #1032, opened 2024-04-08
Update redirects for configuration

sfc-gh-dmatthews:
## 📚 Context
This is a follow-up to the recent IA rework. #1025 

Since `develop/concepts/configuration/options` is more of a stub until it gets rewritten, this PR changes the redirect from the old configuration page to `develop/api-reference/configuration/config.toml`.

Additionally, a historical redirect was corrected to make it consistent with versioned redirects of the same page.


snehankekre (approved):
Small, but important change until the stub is rewritten. Thanks for this!

PR #1033, opened 2024-04-09
Add fragment tutorials and release follow ups

sfc-gh-dmatthews:
## 📚 Context
Fragments were released with Streamlit version 1.33.0.

## 🧠 Description of Changes
Added three new tutorials for fragments.

## 💥 Impact
Size:
- [X] Not small

## 🌐 References
Notion task: [Docs for st.experimental_fragment](https://www.notion.so/snowflake-corp/Docs-for-4e14ec59ab1a4f4ca24c335907d83e5e?pvs=4)
Notion task: [Fragment tutorials](https://www.notion.so/snowflake-corp/Fragments-tutorials-b4e4ca8f507547dfb7c2375f4e142d88?pvs=4)

**Contribution License Agreement**
By submitting this pull request you agree that all contributions to this project are made under the Apache 2.0 license.

PR #1034, opened 2024-04-11
Add fragment tutorial embedded apps

sfc-gh-dmatthews:
## 📚 Context
Add embedded app source code to coordinate with fragment tutorials in #1033 

### DIFF
--- a/python/api-examples-source/tutorials/execution-flow/fragments/tutorial-fragment-multiple-container.py
+++ b/python/api-examples-source/tutorials/execution-flow/fragments/tutorial-fragment-multiple-container.py
@@ -1,0 +1,56 @@
+import streamlit as st
+import time
+
+st.title("Cats!")
+
+row1 = st.columns(3)
+row2 = st.columns(3)
+
+grid = [col.container(height=200) for col in row1 + row2]
+safe_grid = [card.empty() for card in grid]
+
+
+def black_cats():
+    time.sleep(1)
+    st.title("🐈‍⬛ 🐈‍⬛")
+    st.markdown("🐾 🐾 🐾 🐾")
+
+
+def orange_cats():
+    time.sleep(1)
+    st.title("🐈 🐈")
+    st.markdown("🐾 🐾 🐾 🐾")
+
+
+@st.experimental_fragment
+def herd_black_cats(card_a, card_b, card_c):
+    st.button("Herd the black cats")
+    container_a = card_a.container()
+    container_b = card_b.container()
+    container_c = card_c.container()
+    with container_a:
+        black_cats()
+    with container_b:
+        black_cats()
+    with container_c:
+        black_cats()
+
+
+@st.experimental_fragment
+def herd_orange_cats(card_a, card_b, card_c):
+    st.button("Herd the orange cats")
+    container_a = card_a.container()
+    container_b = card_b.container()
+    container_c = card_c.container()
+    with container_a:
+        orange_cats()
+    with container_b:
+        orange_cats()
+    with container_c:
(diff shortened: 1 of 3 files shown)


python/api-examples-source/tutorials/execution-flow/fragments/tutorial-fragment-multiple-container.py
```diff
@@ -0,0 +1,56 @@
...
+    st.button("Herd the black cats")
+    container_a = card_a.container()
+    container_b = card_b.container()
+    container_c = card_c.container()
+    with container_a:
+        black_cats()
+    with container_b:
+        black_cats()
+    with container_c:
+        black_cats()
```
sfc-gh-jcarroll, line 36: Is it intentional to do the extra steps rather than e.g.

```python
with card_a.container():
    black_cats()
```

(I think that's equivalent but if I'm getting that wrong, that would be a good reason lol)

sfc-gh-jcarroll: Ah, if you think the explanation is clearer with the way you have it currently, that's fine w me

python/api-examples-source/tutorials/execution-flow/fragments/tutorial-fragment-rerun.py
```diff
@@ -0,0 +1,87 @@
...
+    product_names = ["Widget " + letter for letter in string.ascii_uppercase]
+    average_daily_sales = np.random.normal(1_000, 300, len(product_names))
+    products = dict(zip(product_names, average_daily_sales))
+
+    data = pd.DataFrame({})
+    sales_dates = np.arange(date(2023, 1, 1), date(2024, 1, 1), timedelta(days=1))
+    for product, sales in products.items():
+        data[product] = np.random.normal(sales, 300, len(sales_dates)).round(2)
+    data.index = sales_dates
+    data.index = data.index.date
```
sfc-gh-jcarroll, line 22: Sorry I know it's a super nit and I'm bad at pandas, is this functionally different from just `data.index = sales_dates.date`? I saw the comment about stripping the timestamp but I was still left slightly confused about the double assignment.

sfc-gh-jcarroll (approved):
Left a couple nits, LGTM once those are addressed or overruled

python/api-examples-source/tutorials/execution-flow/fragments/tutorial-fragment-multiple-container.py
sfc-gh-dmatthews: This is the most protective against undesired behavior when combined with empties. By defining the containers all first, a slow process prevents them from blanking out during a full rerun (which may or may not be desirable). The way it is has the ghosting/resolidifying effect on a full rerun. The compact form you suggested will clear the cards since there's a (possible) delay between creating the empty and the plain container.

(I also need to do some further tests and possibly report another fragment bug. I came across some unexpected behavior with some very subtle variations involving empties...)

python/api-examples-source/tutorials/execution-flow/fragments/tutorial-fragment-rerun.py
sfc-gh-dmatthews: It will error if I try to use `.date` on the `numpy` object. Even though its defined with `date` instead of `dateime`, `pandas` insists on adding the zero timestamps, so (as far as I know) I have to clean it up after creating the `pandas` index object, but there may very well be something cleaner. I figured the `.date` on the index was a nice enough hint that would be more broadly applicable anyway.

PR #1055, opened 2024-05-04
Fix search indexing for html commands and changelog

### LINKED ISSUE #1052: Search st.html goes to page not found
### Summary

When searching for `st.html` and clicking on the result, it goes to **Page not found**, because the URL is wrong. The correct URL must have `.html` at the end.

### Steps to reproduce

1. Go to https://docs.streamlit.io
2. Click on the search bar
3. Type st.html
4. Select the first result

**Screen recording:**

Kapture 2024-05-03 at 11 22 06

**Expected behavior:**

Goes to: https://docs.streamlit.io/develop/api-reference/utilities/st.html

**Actual behavior:**

Goes to: https://docs.streamlit.io/develop/api-reference/utilities/st
(it's missing the .html at the end)

sfc-gh-dmatthews:
## 📚 Context
Streamlit commands ending in `.html` were not indexed correctly (#1052). Additionally, the changelog was not getting indexed because it was too long. This PR fixes that.

## 🧠 Description of Changes
* The search index build script uses `match` with a regex expression instead of `split(".html")` which resulted in incorrect results for pages ending in "...html.html".
* The changelog has been broken apart. The main changelog page only contains those versions which we still document. The rest are moved to an archival page which isn't indexed by Algolia, but is linked from the main changelog.
* Color was added to the search index build script so search results can be color coded to match their categories.

## 🌐 References
Notion project: [Algolia search problems](https://www.notion.so/snowflake-corp/Algolia-search-problems-4eed9ab95d694ab9bc8a628942a59b6d?pvs=4)

Closed #1052 


sfc-gh-dmatthews:
@sfc-gh-tteixeira I went into the search index and manually added a color for the "Quick reference" page. It won't change how the production site displays anything since it's unused, but it let's you see a hypothetical category color incorporated in this preview:

image

From pattern-matching (since I'm not sure the best way to test-run the script), I'm guessing we can pull the colors in the same way we pull the icons. In this way, search results can be color-coded to match their categories. There are two different CSS files for the search modal, so the highlight color is suboptimal as it is less direct (using an `<em>` child of a classed container rather than the direct highlight class applied to `<em>`). I wasn't quite sure if there was a good way to mix classes between to the files... What I really want is something like `.IndigoHighlight ais-Snippet-highlighted`, but I don't know if there's a trick to do that.

sfc-gh-tteixeira (approved):

PR #1124, opened 2024-07-26
Reduce release notes file size

sfc-gh-dmatthews:
## 📚 Context
Algolia fails to update when the release notes page gets too big. This PR moves release notes 1.16 through 1.21 to a "historical release notes" page which is not indexed by Algolia. The versions are still selectable in the API reference.


snehankekre (approved):
LGTM 👍 

Just for my understanding: do we know the exact error (if any) that Algolia throws when the page gets too big (i.e. how do we know that Algolia fails to update)? I want to see if there's a way to increase the limit and/or why Algolia is designed this way.

PR #1132, opened 2024-08-10
Fix scroll padding, so headers don't appear behind the top bar when you navigate to a page via an #anchor.

sfc-gh-tteixeira:
## 📚 Context

When you click navigate to a page like <https://docs.streamlit.io/develop/api-reference/utilities/st.context#contextcookies>, which has a hash pointing to a specific part of the page (`#contextcookies` in this case), that part of the page gets partially obscured by the site's top bar. With this change, this no longer happens.

Compare the above with <https://deploy-preview-1132--streamlit-docs.netlify.app/develop/api-reference/utilities/st.context#contextcookies>! Here you'll the "context cookies" header is visible rather than hidden under the top bar.

## 🧠 Description of Changes

Fix the number used in scroll padding to match what we use for the top bar height.

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

None

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -3,7 +3,7 @@
 @tailwind utilities;
 
 html {
-  @apply overflow-x-hidden overflow-y-auto scroll-pt-12;
+  @apply overflow-x-hidden overflow-y-auto scroll-pt-24;
 }
 
 body {


sfc-gh-dmatthews:
@sfc-gh-tteixeira Thanks! This is another nit I can check off my list. Just FYI, this PR ends up adding excessive padding in other areas. I think it's still better than what it was since at least no header is hidden now, but we might be able to target the scroll padding a little more. See [preview](https://deploy-preview-1132--streamlit-docs.netlify.app/develop/concepts/architecture/fragments#defining-and-calling-a-fragment) vs [current](https://docs.streamlit.io/develop/concepts/architecture/fragments#defining-and-calling-a-fragment).

sfc-gh-dmatthews (approved):

PR #1150, opened 2024-09-14
Allow diff line highlighting for codeblocks

sfc-gh-dmatthews:
## 📚 Context
Add red and green line highlighting for code blocks with language diff. This is helpful for instructions and tutorials when directing users to make edits.

New:
image

Previous:
image

### DIFF
--- a/components/blocks/code.module.css
+++ b/components/blocks/code.module.css
@@ -75,6 +75,13 @@
   @apply inline;
 }
 
+.Pre code :global(.deleted) {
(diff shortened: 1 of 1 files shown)


sfc-gh-dmatthews:
If we can easily the code blocks accept two languages (diff and Python), that would be desirable to keep the text color from Python and only add background color for diff, but it looked like that would take a little more work. This is a five-minute easy solution.

Issue #87, opened 2019-09-10
Add documentation about how to use Streamlit with multiple threads.


tvst:
TLDR: you either need to use the ReportThread class for your thread, or call add_report_ctx on your thread after creation

mansidak:
@tvst @randyzwitch I still can't figure out a clear way to import Report Threads. Like others have said, using `from streamlit.ReportThread import add_report_ctx` only leads to an error saying that this module doesn't exist :/

dpinol:
> @tvst @randyzwitch I still can't figure out a clear way to import Report Threads. Like others have said, using `from streamlit.ReportThread import add_report_ctx` only leads to an error saying that this module doesn't exist :/

Now you need to use from `streamlit.runtime.scriptrunner import add_script_run_ctx` https://github.com/streamlit/streamlit/issues/1326

MoritzNekolla:
`"TLDR: you either need to use the ReportThread class for your thread, or call add_report_ctx on your thread after creation"`

But why is the code snippet below not working. Could you explain how to get that to run @tvst or @dpinol ?
```python
import time
import streamlit as st
from threading import Thread
from streamlit.runtime.scriptrunner import add_script_run_ctx

def target():
    time.sleep(1)
    st.text("thread")

t = Thread(target=target)
add_script_run_ctx(t)
t.start()
```

VEADER1005:
@MoritzNekolla Why am I getting "TypeError: Protocols cannot be instantiated" any clue?

MoritzNekolla:
`Why am I getting "TypeError: Protocols cannot be instantiated" any clue?`

This is caused by python 3.9.7. Try upgrad/downgrade your python version.

jokester:
What should be in the document then? I have a few arguable points though:

1. explain how is user's code executed in absence of custom threads (if it's not introduced elsewhere)
2. `ScriptRunContext`, its lifetime, and the `missing ScriptRunContext` warning
3. how to do custom threading without `ScriptRunContext`
4. how to do custom threading with `ScriptRunContext`

3 and 4 are separated because I don't think `add_script_run_ctx(my_thread)` is the go-to fix for `missing ScriptRunContext`. If I read the code correctly a `ScriptRunContext` is created for a script run, and contains session-specific binding like `UserInfo` `UploadedFileManager`. Using it indifferently can cause memory leak or security issue.

msabramo:
[...]

PR #1154, opened 2024-09-17
add doc to explain multithreading

jokester:
## Warning for interested

This PR likely requires major change before merged, if it ever will. The patterns mentioned should mostly work for current Streamlit, but the official team has the decide to how to introduce them.

## 📚 Context

Surprising many people are confused by multithreading in Streamlit, like in https://github.com/streamlit/docs/issues/87 / https://github.com/streamlit/streamlit/issues/1326 / https://github.com/streamlit/streamlit/issues/8490 .

This PR tries to explain the underneath Streamlit-specific things, and provide multithreading patterns to start with.

## 🧠 Description of Changes

- Add a page to explain Streamlit-created and user-created threads, and how to use them together.

**Revised:**

see the new file

**Current:**

(I didn't find document explaining threads and ScriptRunContext, except in the code)

## 💥 Impact

Size:

- [ ] Small 
- [x] Not small 

## 🌐 References

- https://github.com/streamlit/docs/issues/87
- https://github.com/streamlit/streamlit/issues/1326
- https://github.com/streamlit/streamlit/issues/8490


jokester:
I'm happy to hear feedback about content or format :pray:

jokester:
CI fails but I can't find the error message. The links like https://app.netlify.com/sites/streamlit-docs/deploys/66eaa4a1b1871d000843e854 only says "A fatal rendering error has occurred"

sfc-gh-dmatthews:
Hi @jokester. Thanks for contributing. The site won't build and the page won't be viewable without correctly updating site navigation. Check out the readme file for docs, specifically [Add pages to the menu](https://github.com/streamlit/docs?tab=readme-ov-file#add-pages-to-the-menu).

I'm going over items for the next release right now, so it might take me a few weeks to look at this more closely. Multithreading has been on my list for a while, so thanks for taking a stab at it. I'll circle back here when I get the next release fleshed out. 😄

content/develop/concepts/architecture/threading.md
```diff
@@ -0,0 +1,167 @@
...
+    result_1.write(t1.return_value)
+    result_2.write(t2.return_value)
+
+main()
+
+```
+
+### 2. Expose context object to custom thread
+
+Alternatively, one can let a custom thread have access to the `ScriptRunContext` attached to ScriptThread. This pattern is also used by Streamlit standard widgets like [st.spinner](https://github.com/streamlit/streamlit/blob/develop/lib/streamlit/elements/spinner.py).
```
jokester, line 122: Though I saw people are already going this way (in various GH issues), I'm not really sure about this pattern

1. it exposes internal object to page writers

2. it it less guaranteed. I don't know enough to say the probability where a `ScriptRunContext` suffices. The other pattern just looks "safer" to me because it assumes less.

jokester:
@sfc-gh-dmatthews Thanks !! I managed to get preview build to work https://deploy-preview-1154--streamlit-docs.netlify.app/develop/concepts/architecture/threading

content/develop/concepts/architecture/threading.md
sfc-gh-dmatthews: Officially, adding your own threads isn't supported. We don't want to include unsupported patterns in the main concepts area. This will likely go out as a Knowledge Base (KB) article. There is a subset of this information that can live in concepts, but it will need to be very carefully separated. For now, we can move this page to the KB so that we can get it published faster and I'll likely follow up with moving some of it back into concepts. (I still haven't read through any of it yet, so I still expect a few weeks before I get to this. Just a heads up about what will likely happen.)

In the longer run, the plan is to properly support multithreading and async tasks, but that work isn't currently on the schedule this quarter or next, so it wouldn't likely happen until next year at the earliest.

jokester: That makes sense. I don't really want to promote the hack either.

When you come back, feel free to change or move things or ask me to do so 👍🏽

sfc-gh-dmatthews:
Sorry for the delay. I've been working on this this week. I think instead of the pseudo code to explain the threading structure, an illustration might be better. I have a much more detailed illustration I want to put into a contributing wiki-type document, but for this guide, I want to make sure it's as accessible to Python developers as possible (who might not ever touch a Tornado server to deal with WebSockets). I'm currently thinking of something like this to just schematically show script threads tied to script runs and that they have to communicate both to the front end and the server.

image

I should have the PR updated with a first pass by early next week. 😄

content/develop/concepts/app-design/multithreading.md
```diff
@@ -0,0 +1,180 @@
...
+
+This is a simplifed illustration to show how Streamlit works:
+
+![Each user session uses script threads to communicate between the user's front end and the Streamlit server.](/images/concepts/Streamlit-threading.svg)
+
+## `streamlit.errors.NoSessionContext`
+
+Many Streamlit commands, including `st.session_state`, expect to be called from a script thread. When Streamlit is running as expected, such commands use the `ScriptRunContext` attached to the script thread to ensure they work within the intended session and update the correct user's view. When those Streamlit commands can't find any `ScriptRunContext`, they raise a `streamlit.errors.NoSessionContext` exception. Depending on your logger settings, you may also see a console message identifying a thread by name and warning, "missing ScriptRunContext!"
```
sfc-gh-dmatthews, line 29: Should this in fact be "All Streamlit commands?" @lukasmasuch

content/develop/concepts/app-design/multithreading.md
```diff
@@ -0,0 +1,180 @@
...
+<Cloud name="doc-multithreading-no-st-commands-iterative" height="700px" />
+
+### Option 2: Expose `ScriptRunContext` to the thread
+
+If you want to call Streamlit commands from within your custom threads, you must attach the correct `ScriptRunContext` to the thread.
+
+<Warning>
+
+- This is not officially supported and may change in a future version of Streamlit.
+- This may not work with all Streamlit commands.
```
sfc-gh-dmatthews, line 131: Do we know exhaustively which commands these might be, or should this just be a generic warning to cover the possibilities?

content/develop/concepts/app-design/multithreading.md
```diff
@@ -0,0 +1,179 @@
...
+threads = [
+    WorkerThread(delay, container)
+    for delay, container in zip(delays, result_containers)
+]
+for thread in threads:
+    add_script_run_ctx(thread, get_script_run_ctx())
+    thread.start()
+
+for thread in threads:
+    thread.join()
```
sfc-gh-dmatthews, line 174: (Commenting for engineering review later in the week when people are back from the holidays):

Should we be storing the threads in Session State and running a check for threads at the top of the script? And/or disabling widgets when threading? Adding try-except to the Streamlit commands in the threads? Although the script works like this, if a user reruns the app before the page finishes loading, that'd be an issue. Hence this is very fragile, right?

content/develop/concepts/app-design/multithreading.md
lukasmasuch: Hmm, not 100% sure. I believe a lot of commands will not raise a `NoSessionContext` if there is no `ScriptRunContext`. This is to support bare execution (execute a Streamlit app script with pure python). But calling these commands with the context won't do anything.

content/develop/concepts/app-design/multithreading.md
```diff
@@ -0,0 +1,179 @@
...
+- This may not work with all Streamlit commands.
+- Ensure custom threads do not outlive the script thread owning the `ScriptRunContext`. Leaking of `ScriptRunContext` may cause security vulnerabilities, fatal errors, or unexpected behavior.
+
+</Warning>
+
+In the following example, a custom thread with `ScriptRunContext` attached can call `st.write` without a warning.
+
+```python
+import streamlit as st
+from streamlit.runtime.scriptrunner import add_script_run_ctx, get_script_run_ctx
```
lukasmasuch, line 140: nit: Before we release this, we might want to double-check if it might be better to expose `get_script_run_ctx` and `add_script_run_ctx` in a slightly less internal namespace.

sfc-gh-dmatthews: If there's any hope of doing so relatively quickly/easily, that'd be great!

lukasmasuch: I don't know if it's relatively quickly/easily. Probably needs some discussion with the product (cc @jrieke). I think in the best case, we can semi-officially expose it to a stable namespace (e.g., `streamlit.multithreading`) and also put some metrics tracking on these methods. But that probably takes a bit of decision time. It is probably fine in the meantime to add a warning that these methods are purely internal and can change/break with any version update.

sfc-gh-dmatthews: I do have a warning included in the section: https://deploy-preview-1154--streamlit-docs.netlify.app/develop/concepts/design/multithreading#option-2-expose-scriptruncontext-to-the-thread

I'll bring this up in office hours to see if there are any other concerns before publishing. My biggest question is if I should include a little more careful handling of the exposed thread context. The warning states that custom threads should not outlive the script thread from whence they came, but the example doesn't actually do enough to prevent that since it does nothing to handle an interrupted script run.

sfc-gh-jrieke: Yeah definitely needs a bit of discussion and thought about how we do that API. I know Joshua wanted to do it but I never really deeply looked into multithreading so far, so it doesn't really make sense to do something ad-hoc right now. I think it's fine mentioning the internal API in a guide but we should definitely put in a disclaimer that it's an internal, unstable API and will change in the future.

sfc-gh-dmatthews (approved):

PR #1203, opened 2024-12-12
Change inline code styling.

sfc-gh-tteixeira:
## 🧠 Description of Changes

**Revised:**
Screenshot 2024-12-18 at 09 41 24

**Current:**
Screenshot 2024-12-18 at 09 41 38

## 💥 Impact

Size:

- [ ] Small 
- [ ] Not small 

## 🌐 References

- [ ] [Notion](...)

### DIFF
--- a/styles/text.scss
+++ b/styles/text.scss
@@ -118,6 +118,16 @@ tt.docutils.literal {
   @apply text-base;
 }
 
+/* Improve `code` links. Don't use border, use underline instead. */
+p > a:has(tt.docutils.literal) {
+  @apply border-0;
+}
+
+/* Improve `code` links. Don't use border, use underline instead. */
+p > a > tt.docutils.literal {
+  @apply underline;
+}
+
 p a code {
   @apply border-none p-0;
 }
--- a/styles/text.scss
+++ b/styles/text.scss
@@ -112,9 +112,10 @@ p > code,
 li > code,
 /* Inline code blocks in docstrings */
 tt.docutils.literal {
-  @apply border-b px-1 rounded break-words;
-  @apply text-green-80 bg-gray-10 border-gray-30;
-  @apply dark:text-green-40 dark:bg-gray-90 dark:border-gray-100;
+  @apply break-words px-1 rounded;
+  @apply text-green-80 bg-gray-10;
+  @apply dark:text-green-50 dark:bg-gray-90;
+  @apply text-base;
 }
 
 p a code {
--- a/styles/text.scss
+++ b/styles/text.scss
@@ -99,7 +99,7 @@ ul + p {
   @apply mt-8;
 }
 
-/* Focuss styles */
+/* Focus styles */
 a:focus,
 a:focus-visible,
 button:focus,
@@ -113,8 +113,8 @@ li > code,
 /* Inline code blocks in docstrings */
 tt.docutils.literal {
   @apply break-words px-1 rounded;
-  @apply text-green-80 bg-gray-10;
-  @apply dark:text-green-50 dark:bg-gray-90;
+  @apply text-green-80 bg-green-80/5;
+  @apply dark:text-green-50 dark:bg-green-50/10;
   @apply text-base;
 }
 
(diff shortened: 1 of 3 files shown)


sfc-gh-tteixeira:
Example pages to check:
* /develop/api-reference/data/st.dataframe
* /get-started/installation/command-line
* /develop/api-reference/media/st.logo (see link)

styles/text.scss
```diff
@@ -109,8 +109,23 @@ button:focus-visible {
...
+}
+
+/* Improve `code` links. Don't use border, use underline instead. */
+p > a:has(tt.docutils.literal) {
+  @apply border-0;
+}
+
+/* Improve `code` links. Don't use border, use underline instead. */
+p > a > tt.docutils.literal {
+  @apply underline;
```
sfc-gh-dmatthews, line 128: If we use underline, commands with underscores will look like this:
image

Also, the site uses border for links elsewhere, so if we do switch it, we have other styling to correct too. This is what the preview currently shows:

A link in the function description of st.logo:
image

A link in the class description of SQLConnection:
image

Links from the connections concept page (both standard and header-code):
image


The only thing that's been changed is inline code within a parameter description, I think.

sfc-gh-dmatthews: I'd vote for leaving it as a border to be consistent across all links and just modify the color if that's what you want. Though I'd be a little confused by the color scheme if we make it a green underline for code (to always match the text color), then how do we handle the hover effect. We currently hover black -> red for the bottom border on links, but green -> red feels....seasonal. 😅

sfc-gh-tteixeira: For regular links we'll continue to use the existing design. What we're changing here is the design of links inside inline code from autofunctions.

So from your examples, only the SQL Connection example will be impacted by this PR. The rest will stay the same.

sfc-gh-dmatthews: In the connection to data article, `st.connection` is also inline code; it's just not inline code within Autofunction. Do you mean to exclude that as well?

sfc-gh-dmatthews:
Tips, Info, Warning: e.g. https://deploy-preview-1203--streamlit-docs.netlify.app/develop/concepts/multipage-apps/overview
image


So we have inline code is green except:
* Inline code in headers is black
* Inline code with links in MarkDown (not AutoFunction) is black with a bottom border (matches non-inline-code links)
* Inline code with links in AutoFunciton are green underlined

Is that the correct intent?

sfc-gh-dmatthews:
Similarly for deprecation/note blocks within Autofunction, can you confirm if this is how you want the background colors to look?
image
image

styles/text.scss
```diff
@@ -109,8 +109,25 @@ button:focus-visible {
...
+/* Improve `code` links. Don't use border, use underline instead. */
+a:has(tt:only-child),
+a:has(code:only-child) {
+  @apply border-0;
+}
+
+/* Improve `code` links. Don't use border, use underline instead. */
+a > tt:only-child,
+a > code:only-child {
+  @apply underline;
```
sfc-gh-dmatthews, line 130: Inline code links outside of Autofunction have double underline in default text color:
image

Inline code links in Autofunction (function or class description; not parameter description) have a double underline (underline is green, border is default text color)
image

sfc-gh-tteixeira: Great finds! Should be fixed now.

sfc-gh-dmatthews:
This link contains inline code but isn't entirely inline code. Do we just declare that I can't do that anymore since we're using different underlines for code and non-code?

image

sfc-gh-tteixeira:
No, that should be allowed. The problem here seems to be that `:only-child` doesn't count text nodes as children -- weird! So we'll have to look for another solution...

sfc-gh-tteixeira:
Ok, I tried different solutions but nothing look good. Reverting to the way links used to look.

I don't love it, but it's the best we can do in short notice.

Screenshot 2024-12-17 at 15 53 43

sfc-gh-dmatthews (approved):

PR #1257, opened 2025-03-12
Replace broken link

philipnye:
## 📚 Context

Replace broken link

## 🧠 Description of Changes

- Current link gives a 404
- The current article looks to have moved to https://medium.com/towards-data-science/deploying-a-streamlit-web-app-with-azure-app-service-1f09a2159743
- This is paywalled so I've suggested a replacement article that isn't

Size:

- [x] Small 
- [ ] Not small 


sfc-gh-dmatthews:
Thanks! Sorry for the late response. I've gone ahead and removed the links so we don't send people to a paywall as you've mentioned, but since the link seems to be a bit informal, I opened an issue to write a replacement for it instead. #1292 Thanks again!

PR #1286, opened 2025-05-02
brief asyncio / multiprocessing, remove link to external asyncio guide

jokester:
## 📚 Context

The medium article linked is IMO inappropriate to be referred to from streamlit doc, because its analysis and workaround to the asyncio hurdle is too arguable:

1. Tornado server has a event loop, this does not mean it should be used to run user's (potentially blocking) coroutine.
2. There are apparently other possibilities than `run_until_complete`, like in https://github.com/streamlit/streamlit/issues/8488#issuecomment-2844134343 . I think the author assumed too much on the problem to solve.
3. The workaround it suggested, to monkey patch asyncio module with a library, is too intrusive. 

## 🧠 Description of Changes

1. a little more comparison about multiprocessing / asyncio in this page
2. remove the link to the article.

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 

## 🌐 References

- [ ] [Notion](...)

### DIFF
--- a/content/develop/concepts/app-design/multithreading.md
+++ b/content/develop/concepts/app-design/multithreading.md
@@ -15,7 +15,13 @@ Multithreading is a type of concurrency, which improves the efficiency of comput
 
 Multithreading is just one type of concurrency. Multiprocessing and coroutines are other forms of concurrency. You need to understand how your code is bottlenecked to choose the correct kind of concurrency.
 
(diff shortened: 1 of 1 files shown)


sfc-gh-dmatthews:
@jokester Thanks for the contribution. I'm checking with one of our engineers to check make sure I understand things.

Just fyi, the plan is to introduce additional conceptual pages to cover concurrency broadly and have similar pages for the other types (as opposed to this one which is aimed just at multithreading).

jokester:
Sure 👍🏽 . Concurrency is a bigger and impactful topic.

PR #1294, opened 2025-05-14
Clarify custom components use case

### LINKED ISSUE #1279: Describe how to customize an existing component
**Link to doc page in question (if any):**
https://docs.streamlit.io/develop/concepts/custom-components/create

**Name of the Streamlit feature whose docs need improvement:**
custom components

**What you think the docs should say:**
The docs mention you can create "Custom versions of existing Streamlit elements and widgets, such as st.slider or st.file_uploader."
but I have only found examples of creating components from scratch. 

How can I 'subclass' or modify existing components as described in this bullet? For instance, I want to customize the segmented_control such that each option has a different color. I would like to modify streamlit's segmented control rather than build it from scratch, as my attempts to build from scratch resulted in very unreliable/unpredictable behavior of the interactive elements.

sfc-gh-dmatthews:
## 📚 Context
The PR clarifies a line in the custom components guide.

New:
> Create your own components to use in place of existing Streamlit elements and widgets.

Previous:
> Custom versions of existing Streamlit elements and widgets, such as `st.slider` or `st.file_uploader`

This avoids confusion where someone might think there is an easy way to extend a built-in component. Additionally, the other items in the list were shifted to verb-first style. `__repr_html__` is not longer referenced since we have `st.html` and `st.write` that nicely handle things with HTML representations already.

closes #1279

### DIFF
--- a/content/develop/concepts/custom-components/create-component.md
+++ b/content/develop/concepts/custom-components/create-component.md
@@ -19,7 +19,7 @@ Streamlit Components let you expand the functionality provided in the base Strea
 
 **Types of Streamlit Components you could create include:**
 
-- Custom versions of existing Streamlit elements and widgets, such as `st.slider` or `st.file_uploader`.
+- Create your own components to use in place of existing Streamlit elements and widgets.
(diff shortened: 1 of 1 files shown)


sfc-gh-bnisco (approved):
Thank you!

PR #1324, opened 2025-08-15
Make llms-full autogen on deploy

sfc-gh-tteixeira:
## 📚 Context

`docs.streamlit.io/llms-full.txt` needs to be rebuilt every time the site is modified. Today that's done with `make export`. But it turns out that Netlify doesn't run `make export`, and instead it just calls `npm run export` directly. So this PR addresses that.

## 🧠 Description of Changes

Make Netlify call `make export` when building the site. This is being done in `netlify.toml` rather than in the Netlify GUI because the GUI was complaining. See video below.

https://github.com/user-attachments/assets/da2dce68-e2e9-4872-ae2a-e8e523ca2aab

## 💥 Impact

Size:

- [x] Small 
- [ ] Not small 


netlify.toml
```diff
@@ -1,14 +1,18 @@
+[build]
+command = "npm export"
```
sfc-gh-dmatthews, line 2: Is this supposed to be `npm run export`?

sfc-gh-dmatthews:
Can we delete both llm context files in this PR so they don't clutter up the repo? (llm-full.txt and content_catalog.json)

netlify.toml
sfc-gh-tteixeira: Ooops, it's `make export`. Fixed.

sfc-gh-dmatthews:
Since we have to get uv into the environment to fix the build errors, should we try bumping the runtime while we're at it? It's still on Python 3.8.

sfc-gh-tteixeira:
That sounds like a different PR to me. It's best to keep PRs on topic.

sfc-gh-dmatthews (approved):
I added the uv dependency and checked that the llms_full file is available in the preview.

PR #1490, opened 2026-06-17
Add parallel fragments description

pplettner:
## 📚 Context

Added a whole section in "Working with fragments" describing the behaviour of parallel fragments.

### DIFF
--- a/content/develop/concepts/architecture/fragments.md
+++ b/content/develop/concepts/architecture/fragments.md
@@ -18,6 +18,7 @@ Fragments are versatile and applicable to a wide variety of circumstances. Here 
 - Your app has multiple visualizations and each one takes time to load, but you have a filter input that only updates one of them.
 - You have a dynamic form that doesn't need to update the rest of your app (until the form is complete).
 - You want to automatically update a single component or group of components to stream data.
+- Your app has several slow, independent operations (like database queries or API calls) that you want to run at the same time instead of one after another.
 
 ## Defining and calling a fragment
 
@@ -77,6 +78,8 @@ If you run the code above, the full script will run top to bottom on your app's 
 
 ![Diagram of fragment execution flow](/images/concepts/fragment_diagram.png)
 
+By default, fragments run inline on the main thread, in the order you call them, just like the rest of your script. If you have slow, independent fragments, you can opt in to running them concurrently during full-app reruns with `parallel=True`. See [Run fragments in parallel](#run-fragments-in-parallel) below.
+
 ## Fragment return values and interacting with the rest of your app
 
 Streamlit ignores fragment return values during fragment reruns, so defining return values for your fragment functions is not recommended. Instead, if your fragment needs to share data with the rest of your app, use Session State. Fragments are just functions in your script, so they can access Session State, imported modules, and other Streamlit elements like containers. If your fragment writes to any container created outside of itself, note the following difference in behavior:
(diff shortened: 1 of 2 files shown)


content/develop/concepts/architecture/fragments.md
```diff
@@ -89,6 +92,53 @@ To prevent elements from accumulating in outside containers, use [`st.empty`](/d
...
+    data = another_expensive_query()
+    st.dataframe(data)
+
+slow_chart()
+slow_table()
+```
+
+When `parallel=True`, the behavior depends on the type of rerun:
+
+- **During a full-app rerun**, Streamlit dispatches the fragment to a thread pool. It runs concurrently with your other parallel fragments and with the rest of your main script, rather than blocking on each fragment in turn. If `slow_chart` and `slow_table` each take two seconds, running them in parallel lets your app finish in about two seconds instead of four.
```
sfc-gh-lwilby-1: [nit] I am not super familiar with the formatting in the docs repo, but I am wondering if this should be unbolded? We should keep it consistent with the rest of the docs.

content/develop/concepts/architecture/fragments.md
```diff
@@ -89,6 +92,53 @@ To prevent elements from accumulating in outside containers, use [`st.empty`](/d
...
+
+- **During a fragment rerun** (when a user interacts with a widget inside the fragment), execution stays sequential. The fragment runs by itself on the main thread, exactly like a non-parallel fragment. This keeps your state updates predictable when a user is actively interacting with a fragment.
+
+Parallel fragments are most helpful when each fragment does independent, time-consuming work. If your fragments are fast or depend on each other's results, running them in parallel adds complexity without a meaningful speedup.
```
sfc-gh-lwilby-1: [suggestion] I think if they depend on each other's results it won't be possible to design the code with parallel fragments, perhaps we should drop that, or clarify that there is no mechanism to sync content between parallel fragments.

content/develop/concepts/architecture/fragments.md
```diff
@@ -89,6 +92,53 @@ To prevent elements from accumulating in outside containers, use [`st.empty`](/d
...
+
+### Restricted commands during parallel execution
+
+Because parallel fragments run concurrently on separate threads during the initial (full-app) run, a few Streamlit commands aren't safe to call from inside them and will raise an error. These include:
+
+- [`st.dialog`](/develop/api-reference/execution-flow/st.dialog)
+- [`st.switch_page`](/develop/api-reference/navigation/st.switch_page)
+- Writing to containers created outside the fragment.
+
+These commands work normally during a fragment rerun (for example, after a user interacts with a widget inside the fragment), because fragment reruns are sequential. If you need one of these commands, call it from a non-parallel fragment or from the main body of your script.
```
sfc-gh-lwilby-1: [suggestion] Maybe we should give an example of how to run these from a fragment rerun. Basically, you can conditionally open a dialog based on a user interaction with the fragment (that would trigger a fragment rerun) but not open the dialog unconditionally in the fragment.

content/develop/concepts/architecture/fragments.md
pplettner: You're right, I've removed the bold part to keep the style consistent.

content/develop/concepts/architecture/fragments.md
pplettner: I just removed this section, it doesn't add much to the explanation.

content/develop/concepts/architecture/fragments.md
pplettner: I included a small code snippet explaining how to use them in a fragment rerun.

sfc-gh-lwilby-1 (approved):
LGTM

PR #1494, opened 2026-07-07
Parallel fragments tutorial

pplettner:
## 📚 Context

Adding a new tutorial for the parallel fragments feature

### DIFF
--- a/content/develop/tutorials/execution-flow/fragments/run-fragments-in-parallel.md
+++ b/content/develop/tutorials/execution-flow/fragments/run-fragments-in-parallel.md
@@ -67,8 +67,6 @@ def load_inventory():
     st.session_state.inventory = result
     st.metric("Inventory", result["value"])
 
-start = time.time()
-
 cols = st.columns(3)
 with cols[0]:
     load_sales()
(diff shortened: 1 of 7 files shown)


content/develop/tutorials/execution-flow/fragments/run-fragments-in-parallel.md
```diff
@@ -0,0 +1,187 @@
...
+  ```text
+  streamlit>=1.58.0
+  ```
+
+- You should have a clean working directory called `your-repository`.
+- You should have a basic understanding of fragments and Session State.
+
+## Summary
+
+In this example, you'll build an app that loads data from three independent, slow sources. Without parallel fragments, the app waits for each source in turn, so the total load time is the sum of all three. By marking each source's fragment with `parallel=True`, the three loads run concurrently during a full-app rerun, and the app finishes in about the time of the slowest single source.
```
sfc-gh-lwilby-1, line 36: It could be helpful for users to give some examples of what is meant by independent versus not independent to help guide good app design.

content/develop/tutorials/execution-flow/fragments/run-fragments-in-parallel.md
sfc-gh-lwilby-1, line 1: One thing that could be added is how to use session state along with parallel fragments since there is a potential for shared memory. A separate lock has to be added by the user to make it thread safe.

